PR #31, opened 2024-08-01
feat: aggregation for multiblock proofs

zobront:
This PR implements an `aggregation-client`, which takes in a set of multiblock proofs and verifies them.
- Specifically, it verifies a batch of multi-block proofs and confirms that they're linked. It also takes in a chain of headers as input to confirm there's a valid set of headers that link the last `l1_head` to the first `l1_head`.

It also implements an `agg.rs` bin in `zkvm-host`, which takes in a range of proofs (by name), extracts the public values, and proves the aggregated proof.


aggregation-client/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+    for header in headers.iter() {
+        assert_eq!(header.parent_hash, current_hash, "Header chain broken");
+        current_hash = header.hash_slow();
+    }
+
+    // Verify that the last header's hash matches the L1 head of the last boot info
+    assert_eq!(
+        current_hash, last_boot_info.l1_head,
+        "Last header doesn't match last boot info"
+    );
```
zobront, line 65: I think we need logic here that confirms that the L1 Head of each subproof is included somewhere in this chain of Headers.

zkvm-host/bin/agg.rs
```diff
@@ -0,0 +1,119 @@
...
+    utils::setup_logger();
+
+    let args = Args::parse();
+    let fetcher = SP1KonaDataFetcher::new();
+    let prover = ProverClient::new();
+
+    let (proofs, boot_infos) = load_aggregation_proof_data(args.proofs);
+
+    // Fetch the headers from the L1 head of the last block to the L1 head of the first block.
+    let first_head = boot_infos.last().unwrap().l1_head;
```
zobront, line 70: Instead of defaulting to starting at the last `boot_infos` L1 Head, it would be great if we could pass an additional L1 Head as a CLI argument.

This would make it way easier on the proposer side, because we need to checkpoint whatever L1 Head we're going to use for this proof, and we won't know to do that until we know we're ready to do the aggregation proof.

And a later header will still be able to be walked back to everything we need.

zkvm-host/bin/agg.rs
```diff
@@ -0,0 +1,123 @@
...
+    // Fetch the headers from the L1 head of the first boot info to the current L1 head, inclusive.
+    let first_head = boot_infos.last().unwrap().l1_head;
+
+    // Confirm that the headers are in the correct order.
+    let start_header = fetcher
+        .get_header_by_hash(ChainMode::L1, first_head)
+        .await?;
+    // End header is the current L1 head.
+    // TODO: Should this be the finalized header/fetched from a different source?
+    let end_header = fetcher.get_head(ChainMode::L1).await?;
```
zobront, line 78: @ratankaliani Yeah, this needs to be passed in as a part of the call to this function. Basically `op-proposer` will take its view of the current head, checkpoint it (it will have progressed by the time the checkpoint is done), and then pass that L1 Head to its call to the server to make sure the same one is used in the proof.

zobront: We can pass the hash and then do `get_header_by_hash()`.

zkvm-host/bin/agg.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let (_, vkey) = prover.setup(MULTI_BLOCK_ELF);
+
+    let mut stdin = SP1Stdin::new();
+    for proof in proofs {
+        let SP1Proof::Compressed(compressed_proof) = proof else {
+            panic!();
+        };
+        stdin.write_proof(compressed_proof, vkey.vk.clone());
+    }
```
zobront, line 104: I'm a bit unclear if this `vkey` will be used, or if it's the hardcoded one in the client?

zkvm-host/bin/agg.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    // Write the aggregation inputs to the stdin.
+    stdin.write(&AggregationInputs {
+        boot_infos,
+        headers,
+        l1_head: end_header.hash_slow(),
```
zobront, line 110: Once we pass the final L1 Head in as an arg, we can just use that instead of rehashing.

zkvm-host/bin/agg.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// Start L2 block number.
+    #[arg(short, long, num_args = 1.., value_delimiter = ',')]
+    proofs: Vec<String>,
+
+    #[arg(short, long)]
+    multi: bool,
+
```
zobront, line 22: This argument is never used. I think safe to assume we'll just use agg for multi.

PR #32, opened 2024-08-02
feat: add prove cli flag

zobront:
Add a CLI flag that allows the program to be switched between execute mode and prove mode, which generates and saves a proof.

Note: We no longer use verbosity as it conflicts with the SP1 logger, so I've removed it as a CLI arg.


zobront:
@ratankaliani FYI, the reason I made the verbosity change in the justfile is that setting verbosity that way seems to only set it for the native run, whereas using `RUST_LOG=info` sets it for the zkvm run.

ratankaliani:
@zobront I'm going to deprecate the verbosity flag for the native host execution as we don't seem to run into issues with it anymore. The global init of the tracing subscriber conflicts with how SP1's logger is set up, so to avoid complexity, I'll just remove that option for now as it's unused.

ratankaliani (approved):

PR #33, opened 2024-08-02
feat: get_vk binary

zobront:
When deploying a new chain, we'll want to ensure that the VK used on the chain matches the program we're using. This binary will be added to the deployment process to generate the VK from the ELF and include it in the chain's config.


ratankaliani:
Matt has a PR for this in SP1: https://github.com/succinctlabs/sp1/pull/1274, will close in favor of that.

PR #39, opened 2024-08-10
feat: add bn acceleration

ratankaliani:


ratankaliani:
No longer need this, as it will be added in a different PR.

PR #48, opened 2024-08-14
refactor: `client-programs` + `prover-server`

ratankaliani:
- Move the programs into `client-programs`.
- Create `prover-server` to host server-related functionality.


crates/host-utils/src/fetcher.rs
```diff
@@ -72,6 +73,63 @@ impl SP1KonaDataFetcher {
...
+            if l1_block_header.number < earliest_block_num {
+                earliest_block_num = l1_block_header.number;
+                earliest_l1_header = Some(l1_block_header);
+            }
+        }
+        Ok(earliest_l1_header.unwrap())
+    }
+
+    /// Fetch headers for a range of blocks inclusive.
+    pub async fn fetch_headers_in_range(&self, start: u64, end: u64) -> Result<Vec<Header>> {
```
zobront, line 97: Make sure to `dotenv::dotenv().ok();` here. This was failing for me before because with that missing it was trying to use localhost for L1 RPC.

ratankaliani: Added it when init'ing `SP1KonaDataFetcher`

zkvm-host/bin/server.rs
```diff
@@ -113,7 +112,10 @@ async fn request_agg_proof(
         .map(|proof| proof.proof.clone())
         .collect();
 
-    let headers = fetch_header_preimages(&boot_infos, payload.l1_head).await?;
+    let fetcher = SP1KonaDataFetcher::new();
```
zobront, line 115: The `boot_infos` call above will fail because the public values are abi encoded. We either need to not ABI encode in validity client (here: https://github.com/moongate-forks/kona-sp1/blob/20b0d890b8b18e9974a3692a9862cf72189f7f55/client-programs/validity-client/src/main.rs#L56) or we need to decode as bytes and then into RawBootInfo (like here: https://github.com/moongate-forks/kona-sp1/commit/d6708ffd3c9b124fec8a1d0c21684d0cb5dfe5cf).

zobront: I think the first option is better.

ratankaliani: Fixed this in the Docker PR, but need to do this here too. Will update the program to use `RawBootInfo`, as we don't use the output of `validity-client` on-chain.

ratankaliani: Actually, I'll save the `RawBootInfo` changes till later. I want to confirm the setup works end to end, then will modify the program to do this + make the batch program a pass through.

PR #65, opened 2024-08-19
nits: cleanup ZKL2OutputOracle contract

zobront:


ratankaliani (approved):
Looks good, I'd add back in a specifier above both structs indicating that they're specific to the zkVM context.

PR #69, opened 2024-08-22
feat: custom rollup config

zobront:
Add support for a rollups outside of the `superchain-registry` via a dynamic `RollupConfig`.

### Program Changes
- The `fault-proof` and `range` programs now take in a `BootInfoWithBytes` config that contains the `rollup_config_bytes`. These bytes are hashed inside of the program to get the `rollupConfigHash` which is committed to within the `rollupConfigStruct`. The `RollupConfig` is used in the `BlockExecutor`.
- The aggregation program verifies that the `rollupConfigHash` committed to in each range proof is the same, and the `rollupConfigHash` is then verified to match the one stored in the contract on-chain. This guarantees that only 1 `rollupConfigHash` can be used at a time.

### Script Changes
- Whenever the fetcher is initialized, save the rollup config for the corresponding chain to `rollup-configs/{chain_id}.json`. This path is used during witness generation, so it's important that the relevant rollup configuration is present at that path.
- Modify the `BootInfoStruct` to include the `rollup_config_hash`.

### Contract Changes
- Add a check for the `rollupConfigHash` whenever `proposeL2Output` root is called on `ZKL2OutputOracle`.
- Before reading the `zkl2ooconfig.json`, build & invoke the `fetch-rollup-config` binary, which updates the `vkey`, `startingBlockNumber`, `chainId`, `l2BlockTime`, `rollupConfigHash`, `startingOutputRoot` and `startingTimestamp`. To use a manually set `startingBlockNumber`, set `USE_CACHED_STARTING_BLOCK` in your environment.

### Miscellaneous
- Add support for using the cached starting block in the `zkl2ooconfig.json`, rather than re-initializing it.
- Fix an issue in the cost estimator that didn't handle span batches with the same start and end block.
- Remove unnecessary TODO's and unused documentation from the book.
- Exit the `native_host_runner` correctly.
- Remove the annotation for the `ec_recover` precompile as it's infrequently used.
- Move `.env.example` to the root.

## TODO's
Remaining:

- [ ] Do an E2E test of span and agg proofs with new code. Note that this will require deploying a new contract (with the rollup config hash added) to verify proofs against.
- [X] Update tutorial in other PR to include calling `cargo run --bin fetch_rollup_config` before deploying contracts.


Cargo.toml
```diff
@@ -46,7 +46,7 @@ futures = "0.3.30"
 serde_cbor = "0.11.2"
 log = "0.4.22"
 itertools = "0.13.0"
-
```
ratankaliani, line 49: nit: Remove this

client-programs/fault-proof/src/main.rs
```diff
@@ -39,9 +43,20 @@ fn main() {
             // and in memory oracle.
             if #[cfg(target_os = "zkvm")] {
                 println!("cycle-tracker-start: boot-load");
-                let raw_boot_info = sp1_zkvm::io::read::<RawBootInfo>();
-                sp1_zkvm::io::commit_slice(&raw_boot_info.abi_encode());
-                let boot: Arc<BootInfo> = Arc::new(raw_boot_info.into());
+                let boot_info_with_no_config = sp1_zkvm::io::read::<BootInfoWithNoConfig>();
+                let rollup_config_bytes = sp1_zkvm::io::read_vec();
+                let boot_info_with_hashed_config = BootInfoWithHashedConfig::new(&boot_info_with_no_config, &rollup_config_bytes);
+                sp1_zkvm::io::commit::<BootInfoWithHashedConfig>(&boot_info_with_hashed_config);
```
ratankaliani, line 49: Did you run into issues having a Vec<u8> inside of `BootInfoWithHashedConfig`? You should just be able to read this in as one struct.

client-programs/range/src/main.rs
```diff
@@ -50,9 +50,20 @@ fn main() {
             // and in memory oracle.
             if #[cfg(target_os = "zkvm")] {
                 println!("cycle-tracker-start: boot-load");
-                let boot = sp1_zkvm::io::read::<RawBootInfo>();
-                sp1_zkvm::io::commit_slice(&boot.abi_encode());
-                let boot: Arc<BootInfo> = Arc::new(boot.into());
+                let boot_info_with_no_config = sp1_zkvm::io::read::<BootInfoWithNoConfig>();
+                let rollup_config_bytes = sp1_zkvm::io::read_vec();
+                let boot_info_with_hashed_config = BootInfoWithHashedConfig::new(&boot_info_with_no_config, &rollup_config_bytes);
+                sp1_zkvm::io::commit::<BootInfoWithHashedConfig>(&boot_info_with_hashed_config);
```
ratankaliani, line 56: Same here.

justfile
ratankaliani, line 1: Is there a way to do this in Rust? I'm hesistant to have complex logic in `justfile`'s as it's not as maintainble.

Cargo.toml
zobront: @ratankaliani Remove what?

justfile
zobront: Updated so all logic is in a Rust binary instead. Agreed it's much better, thanks.

client-programs/fault-proof/src/main.rs
zobront: Done.

Cargo.toml
ratankaliani: The `tiny-keccak` patch. It's included at the bottom of the `Cargo.toml` as well.

client-programs/fault-proof/src/main.rs
```diff
@@ -39,9 +43,19 @@ fn main() {
...
             if #[cfg(target_os = "zkvm")] {
                 println!("cycle-tracker-start: boot-load");
-                let raw_boot_info = sp1_zkvm::io::read::<RawBootInfo>();
-                sp1_zkvm::io::commit_slice(&raw_boot_info.abi_encode());
-                let boot: Arc<BootInfo> = Arc::new(raw_boot_info.into());
+                let boot_info_with_bytes_config = sp1_zkvm::io::read::<BootInfoWithBytesConfig>();
+                let boot_info_with_hashed_config = BootInfoWithHashedConfig::new(&boot_info_with_bytes_config);
+                sp1_zkvm::io::commit::<BootInfoWithHashedConfig>(&boot_info_with_hashed_config);
+
+                let rollup_config: RollupConfig = serde_json::from_slice(&boot_info_with_bytes_config.rollup_config_bytes).expect("failed to parse rollup config");
```
ratankaliani, line 50: Has this been tested? Is the config hashed?

PR #86, opened 2024-08-30
chore: create scripts folder

jtguibas:
- Move `native-host` and `zkvm-host` into `scripts` folder.


ratankaliani (commented):
The build scripts need to be modified to point to the new locations. Adding now.

PR #127, opened 2024-09-23
feat: enable single block proving

TomTaehoonKim:
This PR enables single block proving for op-succinct.

Plus, SP1 version is bumped up to `v2.0.0` since the prover network currently only supports v2.X.X and v3.X.X of SP1.


ratankaliani (approved):
LGTM!

Issue #140, opened 2024-10-03
Print Calldata from Scripts


ratankaliani:
Modify the scripts for rollup operators that use multi-sigs to manage their chain deployments.

PR #143, opened 2024-10-05
feat: raas deployment fixes

ratankaliani:
- Update `produce_payload` to match upstream.
- Use the correct RPC in the rollup config.
- Only pass the `rollup_config_path` and not the chain ID.
- Remove failing test from `OPSuccinctL2OutputOracle.t.sol`.
- Output the raw calldata for the transactions in the forge scripts.


ratankaliani:
Resolves https://github.com/succinctlabs/op-succinct/issues/140

PR #201, opened 2024-11-02
feat(contracts): Use `immutable` variables for storage

ratankaliani:
We only initialize storage variables once per implementation deployment. As such, we can use `immutable` variables to save on storage loads.


ratankaliani:
For now, we've decided to table this feature, as the gas cost savings for a `proposeL2Output` call are at maximum ~4-5% and the `immutable` pattern is not recommended by OZ: https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#why-cant-i-use-immutable-variables

PR #214, opened 2024-11-19
fix: `cost-estimator` OOM 

ratankaliani:


ratankaliani:
Included branch of SP1 that has the changes here: https://github.com/succinctlabs/op-succinct/pull/212

PR #215, opened 2024-11-19
docs: fix typo

PinelliaC:
### Description
Encountered an error when running the command:

```bash
error: unexpected argument '--elf' found
```
It seems that the correct argument might be `--elf-name` instead.

see: https://github.com/succinctlabs/sp1/blob/dev/crates/build/src/lib.rs#L70


ratankaliani:
Thanks, included in https://github.com/succinctlabs/op-succinct/pull/284

PR #216, opened 2024-11-19
feat: display cycle count on CI

leruaa:
The goal of this PR is to display stats on a proof execution on CI.

In order to not pollute `multi` binary with CI related stuff, I did a small refactoring of the witness generation and proof execution part, moving them to `lib` so they can be used by the binary and also an integration test.

The code that call Github API is mainly copied from the one in SP1.

I also did a small change to the cost estimator binary to alway take a block range in the last 2 weeks.

### DIFF
--- a/scripts/prove/bin/multi.rs
+++ b/scripts/prove/bin/multi.rs
@@ -106,7 +106,6 @@ async fn main() -> Result<()> {
             &prover,
             &data_fetcher,
             sp1_stdin,
-            l2_chain_id,
             l2_start_block,
             l2_end_block,
         )
@@ -120,6 +119,12 @@ async fn main() -> Result<()> {
         );
 
         println!("Execution Stats: \n{:?}", stats);
+
+        // Create the report directory if it doesn't exist.
+        let report_dir = format!("execution-reports/multi/{}", l2_chain_id);
+        if !std::path::Path::new(&report_dir).exists() {
+            fs::create_dir_all(&report_dir)?;
+        }
 
         let report_path = format!(
             "execution-reports/multi/{}/{}-{}.csv",
--- a/scripts/prove/bin/multi.rs
+++ b/scripts/prove/bin/multi.rs
@@ -7,7 +7,7 @@ use op_succinct_host_utils::{
     stats::ExecutionStats,
     ProgramType,
 };
-use op_succinct_prove::{execute_proof, generate_witness, DEFAULT_RANGE, MULTI_BLOCK_ELF};
+use op_succinct_prove::{execute_multi, generate_witness, DEFAULT_RANGE, MULTI_BLOCK_ELF};
 use sp1_sdk::{utils, ProverClient};
 use std::{fs, time::Duration};
 
@@ -102,7 +102,7 @@ async fn main() -> Result<()> {
     } else {
         let l2_chain_id = data_fetcher.get_l2_chain_id().await?;
 
-        let (block_data, report, execution_duration) = execute_proof(
+        let (block_data, report, execution_duration) = execute_multi(
             &prover,
             &data_fetcher,
             sp1_stdin,
--- a/scripts/prove/bin/multi.rs
+++ b/scripts/prove/bin/multi.rs
@@ -37,7 +37,7 @@ struct Args {
     prove: bool,
 
     /// Env file.
-    #[arg(short, long, default_value = ".env")]
+    #[arg(long, default_value = ".env")]
     env_file: Option<String>,
 }
 
--- a/scripts/prove/bin/multi.rs
+++ b/scripts/prove/bin/multi.rs
@@ -5,13 +5,11 @@ use op_succinct_host_utils::{
     fetcher::{CacheMode, OPSuccinctDataFetcher},
     get_proof_stdin,
     stats::ExecutionStats,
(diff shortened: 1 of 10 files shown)


scripts/prove/bin/multi.rs
```diff
@@ -113,37 +100,32 @@ async fn main() -> Result<()> {
...
-
-        // Create the report directory if it doesn't exist.
-        let report_dir = format!("execution-reports/multi/{}", l2_chain_id);
-        if !std::path::Path::new(&report_dir).exists() {
-            fs::create_dir_all(&report_dir).unwrap();
-        }
-
-        let block_data = data_fetcher
-            .get_l2_block_data_range(l2_start_block, l2_end_block)
-            .await?;
```
ratankaliani, line 137: Why did you remove this? We need to create the directory unless it already exists here.

ratankaliani: nvm

utils/host/src/block_range.rs
```diff
@@ -42,3 +42,17 @@ pub async fn get_validated_block_range(
 
     Ok((l2_start_block, l2_end_block))
 }
+
+// Get a fixed recent (less than 2 weeks) block range.
```
ratankaliani, line 46: nit: Make a Rustdoc

utils/host/src/block_range.rs
```diff
@@ -42,3 +42,17 @@ pub async fn get_validated_block_range(
...
+pub async fn get_rolling_block_range(
+    data_fetcher: &OPSuccinctDataFetcher,
+    range: u64,
+) -> Result<(u64, u64)> {
+    let header = data_fetcher.get_l2_header(BlockId::finalized()).await?;
+    let l2_block_time = data_fetcher.get_l2_block_time().await?;
+    let block_count_in_two_weeks = 14 * 24 * 60 * 60 / l2_block_time;
+    let l2_end_block = (header.number / block_count_in_two_weeks) * block_count_in_two_weeks;
```
ratankaliani, line 54: Why not just use `header.number` here? You're diving by, then multiply by `block_count_in_two_weeks`.

utils/host/src/block_range.rs
ratankaliani, line 46: Use rustdoc `///` for fn comments.

scripts/prove/tests/common/mod.rs
```diff
@@ -0,0 +1,67 @@
+use reqwest::Client;
+use serde_json::json;
+
+pub async fn post_to_github_pr(
```
ratankaliani, line 4: nit: Add fn comment

scripts/prove/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    let witness_generation_time_sec = start_time.elapsed();
+    println!(
+        "Witness Generation Duration: {:?}",
+        witness_generation_time_sec.as_secs()
+    );
+
+    Ok(witness_generation_time_sec)
+}
+
+pub async fn execute_proof(
```
ratankaliani, line 35: `execute_multi` is a better name

scripts/prove/tests/multi.rs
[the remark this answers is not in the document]
ratankaliani: nvm, probably good given that it's CI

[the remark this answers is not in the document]
leruaa: There are a few reasons:

- I didn't wanted to "pollute" the binary with CI stuff (i.e. sending to GitHub API)
- The function that create the comment in GitHub can easily used in other integration tests
- If we add other integration tests, we could run them in paralell
- This IS an integration test ;)

utils/host/src/block_range.rs
```diff
@@ -42,3 +44,17 @@ pub async fn get_validated_block_range(
...
     Ok((l2_start_block, l2_end_block))
 }
+
+/// Get a fixed recent (less than 2 weeks) block range.
+pub async fn get_rolling_block_range(
+    data_fetcher: &OPSuccinctDataFetcher,
+    range: u64,
+) -> Result<(u64, u64)> {
+    let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
+    let start_timestanp = now - (now % (14 * 24 * 60 * 60));
```
ratankaliani, line 54: nit: Pass the "fixed" window (2 weeks) to the function, and use a constant for it inside of the multi/cost estimator script.

scripts/prove/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+        .execute(MULTI_BLOCK_ELF, sp1_stdin.clone())
+        .run()
+        .unwrap();
+    let execution_duration = start_time.elapsed();
+
+    // Create the report directory if it doesn't exist.
+    let report_dir = format!("execution-reports/multi/{}", l2_chain_id);
+    if !std::path::Path::new(&report_dir).exists() {
+        fs::create_dir_all(&report_dir)?;
+    }
```
ratankaliani, line 54: Taste, but this shouldn't occur inside of this function, should instead be inside of the binary/test that's running it, as you materialize the `report_path` separately there as well. Would remove this and move it to the `multi` script.

ratankaliani (approved):
Small refactoring/naming convention changes, otherwise LGTM

PR #218, opened 2024-11-21
fix: post report to GitHub on PRs only

leruaa:
Adds an `POST_TO_GITHUB` env variable to control when report is posted to GitHub.


ratankaliani (approved):
LGTM

PR #225, opened 2024-11-22
fix(fetcher): RPC config

ratankaliani:
This PR introduced a bug: https://github.com/succinctlabs/op-succinct/pull/223, where the URL passed to `kona`'s witness generation had a trailing slash. This PR removes the trailing slash and modifies the fetcher to store the `Url` type for better type-safety.

### DIFF
--- a/utils/host/src/fetcher.rs
+++ b/utils/host/src/fetcher.rs
@@ -54,12 +54,12 @@ impl Default for OPSuccinctDataFetcher {
     }
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct RPCConfig {
-    pub l1_rpc: String,
-    pub l1_beacon_rpc: String,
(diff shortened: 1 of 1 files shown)


utils/host/src/fetcher.rs
```diff
@@ -487,13 +482,13 @@ impl OPSuccinctDataFetcher {
...
 
-    async fn fetch_rpc_data<T>(url: &str, method: &str, params: Vec<Value>) -> Result<T>
+    async fn fetch_rpc_data<T>(url: &Url, method: &str, params: Vec<Value>) -> Result<T>
     where
         T: serde::de::DeserializeOwned,
     {
         let client = reqwest::Client::new();
         let response = client
-            .post(url)
+            .post(url.as_str())
```
leruaa, line 491: The post() parameter is U: IntoUrl so just passing url should work

PR #291, opened 2024-12-17
 feat: add SKIP_L1_PROPOSAL flag to skip L1 transactions

### LINKED ISSUE #258: Add flag to skip L1 transaction
# Description
 It is useful to use op-succinct to generate proofs, but in certain circumstances it isn't required that those proofs are written to the L1. Specifically, during testing it would be great to just be able to run op-succinct in a a mode where we can generate the proofs locally and verify them from the database.
 
 # Feature
Add a new flag `--skip-l1-proposal` (though the name is up to you), which simply skips the L1 transaction step, and instead emits a INFO-level log and still writes the resulting proof information to the database.

crStiv:
# Description
Added functionality to skip L1 transactions during testing by introducing a new environment variable `SKIP_L1_PROPOSAL`.

## Changes
- Added `SKIP_L1_PROPOSAL` environment variable (default: false)
- When `SKIP_L1_PROPOSAL=true`, the script sets `OP_SUCCINCT_SKIP_L1=true`
- Added logging to indicate when L1 transactions are being skipped

## Usage
To skip L1 transactions during testing:
bash
SKIP_L1_PROPOSAL=true ./tools/ci/setup.sh

## Testing
- [ ] Tested with `SKIP_L1_PROPOSAL=false` (default behavior)
- [ ] Tested with `SKIP_L1_PROPOSAL=true`
- [ ] Verified that proofs are still generated and stored in database

Closes #258

### DIFF
--- a/tools/ci/setup.sh
+++ b/tools/ci/setup.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -euo pipefail
+
+# Add default parameter for skipping L1 transactions
+SKIP_L1_PROPOSAL=${SKIP_L1_PROPOSAL:-false}
+
+echo "--- Installing yarn dependencies"
+
+# Disable global cache so that we can cache `.yarn/cache` in buildkite
+yarn config set enableGlobalCache false
+
+# Check for lockfile before running yarn --immutable
+if [[ -f yarn.lock ]]; then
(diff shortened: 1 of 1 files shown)


crStiv:
@ratankaliani hey may i know why did you rejected my pr? maybe ill try to redo it?

PR #292, opened 2024-12-17
feat(contracts): support fallback to optimistic mode

emilianobonassi:


ratankaliani:
Replacing with https://github.com/succinctlabs/op-succinct/pull/293, thanks @emilianobonassi!

PR #303, opened 2024-12-23
feat(proposer): more configurability for witness gen

emilianobonassi:
support configuring more or less than 5 concurrent witness gen and tunable timeout for them


ratankaliani (approved):
LGTM! Thanks for the PR

PR #305, opened 2024-12-30
fix: split midblock block number

ctian1:
When attempting to split a block range due to unfulfillable proof: Block ranges in the DB are `[start, end)`, so start of the second half should be just `midBlock` instead of `midBlock+1`


ratankaliani (approved):
LGTM

PR #321, opened 2025-01-12
feat(contracts): various deployment improvements

emilianobonassi:
Some improvements to facilitate deployments, especially on large scale:

- reuse already deployed logic (maybe deterministically by succinct team per chain), just deploy the proxy on top, important savings ($80)
- support proxyadmin pattern which is the default followed by op stack deployments (proxy onwer is not directly an eoa but a contract - ProxyAdmin)
- optionally use a different key for logic deployment, keeping the admin one just for proxy pointer change


contracts/script/OPSuccinctUpgrader.s.sol
```diff
@@ -9,16 +9,31 @@ import {console} from "forge-std/console.sol";
...
+            console.log("Deploying new logic");
+            OPSuccinctL2OutputOracleImpl = address(new OPSuccinctL2OutputOracle());
+
+            vm.stopBroadcast();
+        }
+
+        vm.startBroadcast(adminPk);
 
-        upgradeAndInitialize(OPSuccinctL2OutputOracleImpl, cfg, l2OutputOracleProxy, executeUpgradeCall);
+        upgradeAndInitialize(OPSuccinctL2OutputOracleImpl, cfg, l2OutputOracleProxy, executeUpgradeCall, proxyAdmin);
```
ratankaliani, line 36: Can you add documentation for these in this section of the book?

https://succinctlabs.github.io/op-succinct/contracts/configuration.html

contracts/script/OPSuccinctDeployer.s.sol
```diff
@@ -17,7 +17,7 @@ contract OPSuccinctDeployer is Script, Utils {
         OPSuccinctL2OutputOracle oracleImpl = new OPSuccinctL2OutputOracle();
         Proxy proxy = new Proxy(msg.sender);
 
-        upgradeAndInitialize(address(oracleImpl), config, address(proxy), true);
+        upgradeAndInitialize(address(oracleImpl), config, address(proxy), true, address(0));
```
ratankaliani, line 20: There's also a case where users may not have an L2OO Proxy but want to re-use the implementation logic that already exists.

E.g. They want to deploy an OP Succinct chain on Mainnet, and there's an implementation contract that's already been deployed.

This forge script should be logic for deploying a new proxy, and should support the `OP_SUCCINCT_L2_OUTPUT_ORACLE_IMPL` varaible.

Can you also configure this script to support `adminPk` and `deployPk`? Currently, it just sets the owner of the Proxy to `msg.sender`, when it should use the account associated with `admin_pk` if it's set.

PR #325, opened 2025-01-14
docs: add troubleshooting.md

fakedev9999:
Adds a new troubleshooting.md file to the op-succinct book, providing guidance on common issues encountered and their solutions.

The doc currently handles two common issues with op-succinct; Missing Trie Node Error and L2 Block Validation Failure.


ratankaliani (approved):
LGTM

PR #334, opened 2025-01-16
chore(utils): update block data range

ratankaliani:
Fixes fetching the block data for the very first block in the range, which is actually not executed nor proven. This prevents including non-executed blocks in stats, especially for the number of blocks, leading to more accurate estimates.

### DIFF
--- a/utils/host/src/stats.rs
+++ b/utils/host/src/stats.rs
@@ -129,7 +129,8 @@ impl ExecutionStats {
         let total_gas_used: u64 = block_data.iter().map(|b| b.gas_used).sum();
 
         Self {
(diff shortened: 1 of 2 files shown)


utils/host/src/stats.rs
```diff
@@ -129,7 +129,8 @@ impl ExecutionStats {
         let total_gas_used: u64 = block_data.iter().map(|b| b.gas_used).sum();
 
         Self {
-            batch_start: block_data[0].block_number,
+            // The first block is not included in the cycles, or in the gas used.
+            batch_start: block_data[0].block_number + 1,
```
fakedev9999, line 133: @ratankaliani You should drop the last commit. `-1` is correct.

PR #336, opened 2025-01-17
fix: enable command to run cost estimator default

fakedev9999:
Enables command to run cost estimator default, which is described in the cost estimator doc.

Includes change to use latest finalized block for end block as default. This is to avoid L2 Block Validation Failure error when
running cost estimator default.


ratankaliani (approved):
LGTM

PR #338, opened 2025-01-17
feat: update OP_SUCCINCT game type

leruaa:
The PR adding the `OP_SUCCINCT` game type [has been merged in optimism](https://github.com/ethereum-optimism/optimism/pull/13780#event-15982558845), but finally the number is `6`, not `3`

### DIFF
--- a/contracts/src/OPSuccinctDisputeGame.sol
+++ b/contracts/src/OPSuccinctDisputeGame.sol
@@ -53,9 +53,9 @@ contract OPSuccinctDisputeGame is ISemver, CWIA, IDisputeGame {
     ///      i.e. The game type should indicate the security model.
     /// @return gameType_ The type of proof system being used.
     function gameType() public pure returns (GameType) {
-        // TODO: Once the following PR https://github.com/ethereum-optimism/optimism/pull/13780 is merged,
+        // TODO: Once the a new version of the Optimism contracts is released,
         // update this to return the correct game type: GameTypes.OP_SUCCINCT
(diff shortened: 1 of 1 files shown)


contracts/src/OPSuccinctDisputeGame.sol
```diff
@@ -53,9 +53,10 @@ contract OPSuccinctDisputeGame is ISemver, CWIA, IDisputeGame {
     ///      i.e. The game type should indicate the security model.
     /// @return gameType_ The type of proof system being used.
     function gameType() public pure returns (GameType) {
-        // TODO: Once the following PR https://github.com/ethereum-optimism/optimism/pull/13780 is merged,
+        // TODO: Once a new version of the Optimism contracts containing the PR below is released,
```
fakedev9999, line 56: 1. Do we need to keep the TODO even though the PR is merged?
2. Would there be a chance that OP_SUCCINCT game type randomly changes in the future?

ratankaliani: We need a released version of the contracts that we can use here. Don't think the type will change now that it will be upstreamed.

ratankaliani (approved):

contracts/src/OPSuccinctDisputeGame.sol
leruaa: Yes exactly

fakedev9999 (approved):
LGTM

PR #356, opened 2025-01-31
feat: proofs API

vcastellm:
In this PR we're introducing a new endpoint in the JSON RPC API, to support requesting already generated proofs from the DB. The purpose of this is to facilitate integration of external managers of proofs, in a way they can decide how to aggregate the more granular span proofs together.

Test the new method using:
`cast rpc proofs_requestAggProof [start_block] [end_block] [l1_block_number] [l1_block_hash] --rpc-url http://localhost:8545
`


docker-compose.yml
```diff
@@ -25,3 +25,9 @@ services:
     # The metrics port is the default port for the OP Proposer.
     ports:
       - "7300:7300"
+      - "8545:8545"
```
vcastellm, line 28: This comes handy

docker-compose.yml
```diff
@@ -25,3 +25,9 @@ services:
...
+
+networks:
```
vcastellm, line 30: not harmful and facilitates working with kurtosis

vcastellm:
superseded by #417

PR #359, opened 2025-02-01
fix(ci): elf

ratankaliani:


fakedev9999:
Closing since it's fixed in #372.

PR #370, opened 2025-02-05
feat(contracts): implement `OPSuccinctPermissionedFaultDisputeGame`

fakedev9999:
The `OPSuccinctPermissionedFaultDisputeGame` is a specialized child contract of the `OPSuccinctFaultDisputeGame` that implements role-based access control to restrict who can participate in the game. This contract can serve as a fallback in cases where a permissionless system is undesirable or if the existing permissionless system encounters issues.

# Overview
In a typical fault proof system, multiple entities may freely propose and challenge state roots in order to reach consensus. However, permissionless designs can introduce complexity and costs that some networks might prefer to avoid. By contrast, the `OPSuccinctPermissionedFaultDisputeGame` restricts participation via two new roles: a Proposer and a Challenger.

Each relevant interaction—including prove, challenge, resolve—is accessible only to addresses that have been granted one of these roles. Furthermore, the initialize() function is exclusively reserved for the Proposer.

In the event that you wish to switch back to a permissionless approach, the `RESPECTED_GAME_TYPE` parameter in the `OptimismPortal` can be changed to switch back.

# Optimism's implementation
See [PermissionedDisputeGame.sol](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/dispute/PermissionedDisputeGame.sol)

### DIFF
--- a/contracts/src/fp/OPSuccinctPermissionedFaultDisputeGame.sol
+++ b/contracts/src/fp/OPSuccinctPermissionedFaultDisputeGame.sol
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import "forge-std/console.sol";
+
+// Contracts
+import {OPSuccinctFaultDisputeGame} from "src/fp/OPSuccinctFaultDisputeGame.sol";
+
+// Libraries
+import {Duration, GameStatus} from "src/dispute/lib/Types.sol";
+import {BadAuth} from "src/dispute/lib/Errors.sol";
+
+// Interfaces
+import {ISP1Verifier} from "@sp1-contracts/src/ISP1Verifier.sol";
+import {IDisputeGameFactory} from "src/dispute/interfaces/IDisputeGameFactory.sol";
+
+/// @title OPSuccinctPermissionedFaultDisputeGame
+/// @notice OPSuccinctPermissionedFaultDisputeGame is a contract that inherits from `OPSuccinctFaultDisputeGame`, and contains two roles:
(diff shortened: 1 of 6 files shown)


contracts/src/fp/OPSuccinctPermissionedFaultDisputeGame.sol
```diff
@@ -0,0 +1,108 @@
...
+        }
+        _;
+    }
+
+    modifier onlyProposer() {
+        if (msg.sender != PROPOSER) {
+            revert BadAuth();
+        }
+        _;
+    }
```
ratankaliani, line 43: One concern I have with this design is that in most practical settings to reach Stage 1, you want to have multiple challengers and proposers. 

Do you think it's better to have a pass-through contract where games can be created where this logic can be validated?

Or do you want to allow passing arrays for both of these arguments?

My worry with this design is that it's not easily extensible.

fakedev9999:
@ratankaliani made a very good point https://github.com/succinctlabs/op-succinct/pull/370#discussion_r1943437700

But in CWIA pattern, we can't have immutable args with types of array or mapping.
One alternative is to have a separate "Access Manager" contract and add another immutable arg `ACCESS_MANAGER_ADDRESS` to the constructor of the game contract and do the clone and initalize pattern.

However, it seems better to have an entrypoint contract that wraps over the Factory since:
- it's easily extensible for permissioned games (we don't need to have a separate contract for permission control, but just an array or a mapping)
- we can easily add the "Bank" contract feature at the same time without having a separate WETH kind of contract.

PR #372, opened 2025-02-06
ci: fix ELF workflow

fakedev9999:
After #344, CI for ELF is failing.

This PR fixes it by removing build scripts from maili, which is a new dep added when bumping up kona.

[Commit](https://github.com/succinctlabs/maili/commit/56a0afb44e417dbb0dcba7e09e281f2a218df4c3) that patches maili.


fakedev9999:
Will be fixed in #375 with minimal reproducible version.

PR #375, opened 2025-02-07
chore: patch maili with pre-release hook removed

fakedev9999:
Patches maili with [pre-release-hook](https://github.com/succinctlabs/maili/blob/d75af236d4c2bff1a04235577c7a5273788b6acd/release.toml#L10C1-L10C90) removed in release.toml, which was breaking reproducible ELFs for op-succinct programs.


fakedev9999:
Merged in op-rs/maili https://github.com/op-rs/maili/pull/210.

PR #377, opened 2025-02-08
chore: remove maili-registry from dep

fakedev9999:
In #375, I tried to fix ELF workflow by removing pre-release-hook which doesn't reason well why it breaks reproducible ELFs. I found build.rs which seems a huge sus in maili-registry. This PR tries to fix the ELF workflow by removing unnecessary dep.


fakedev9999:
ELF workflow is still failing. Will investigate future in a direction related to shell scripts.

PR #386, opened 2025-02-13
feat(contracts): remove bond for validity DGF

fakedev9999:
Bonds are not needed for op-succinct validity DGF. This PR removes proposer bond initialization for DGF deployment.


ratankaliani (approved):
LGTM

PR #388, opened 2025-02-13
Initial `hana` support

Ferret-san:
Adding support for OP Stack x Celestia DA chains by using [hana](https://github.com/celestiaorg/hana)


ratankaliani (commented):
Looking good so far! In addition to modifying the witness generation (host) and the program execution with the pipeline (client), we'll need to also modify the proposer to only request for proofs after the data from Celestia corresponding to the L2 blocks we want to prove has been relayed with Blobstream.

Is there an RPC endpoint to get the safe head of the L2 with respect to Blobstream? i.e. At this current time based on the state of the Blobstream contract on L1, what's the highest L2 block that I can prove with Celestia?

fakedev9999:
Closing since #463 will have the same feature fully functioning!

PR #378, opened 2025-02-10
feat(contracts): use `DisputeGameFactory`

fakedev9999:
# Changes
1. `OPSuccinctFaultDisputeGameFactory` is removed.
2. `OPSuccinctDGFDeployer.s.sol` is updated to use OP's [`DisputeGameFactory`](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.8.0/packages/contracts-bedrock/src/dispute/DisputeGameFactory.sol).
3. Go proposer is updated to 1. use `DisputeGameFactory` ABI and 2. `DISPUTE_GAME_TYPE` flag is added so that we can create games through DisputeGameFactory.

PR #391, opened 2025-02-13
chore: fix contract verification

fakedev9999:
Fixes contract verification error for `DisputeGameFactory`.

When running `just deploy-dispute-game-factory` below verification error occurred. This was due to missing change from #378, which needs `src/universal` remapping added.

```
Fail - Unable to verify. Solidity Compilation Error: Source "src/universal/interfaces/ISemver.sol" not found: File not found.
```


ratankaliani (commented):
Can we just use one remapping for `@optimism`, rather than many?

ratankaliani (approved):
LGTM

PR #349, opened 2025-01-27
feat: implement proposer

fakedev9999:
Implements a new component responsible for creating and managing OP-Succinct fault dispute games on L1. The proposer monitors the L2 chain and creates dispute games at configurable intervals to ensure L2 state transition validity.

## Key Features
- Automated dispute game creation at configurable block intervals
- Optional game resolution for unchallenged games
- Continuous L2 chain monitoring
- Fast finality mode with proof inclusion
- Configurable via environment variables

You can find the proposer creating games on ETH sepolia network: [Factory Contract](https://sepolia.etherscan.io/address/0x4206ef931621151471cAa480a0312BD64C7a5591)

You can also find proposer resolving unchallenged games that timed out on ETH sepolia network: [Resolve tx](https://sepolia.etherscan.io/tx/0x6a1b0eaedb8febb553e9e57ae11d54a3df63b21a0aff163a2eb38e75a5069254)

PR #399, opened 2025-02-19
feat(proposer): rust validity proposer

ratankaliani:
# Overview
Replace the fork of `op-proposer` with a Rust proposer. This unifies the validity binaries and allows just running a single container.

# Features
- Prometheus metrics for the Rust binary.
- Expose `range-elf-embedded` and `range-elf-bump`.
- Remove `ProgramType` and `RunContext`.
- Use `sqlx` for compile-time query checks.
- Better logging on validate config mismatches.
- Nicely handle panics from underlying libraries (e.g. `kona`).


proposer/succinct/bin/validity.rs
```diff
@@ -0,0 +1,21 @@
+use std::env;
+
+use anyhow::Result;
+use op_succinct_proposer::Proposer;
+use sp1_sdk::utils;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    // Enable logging.
+    env::set_var("RUST_LOG", "info");
```
fakedev9999, line 10: nit: IMO below pattern in #349 is better since log level is configurable.
```
use tracing_subscriber::{fmt, EnvFilter};

pub fn setup_logging() {
    let format = fmt::format()
        .with_level(true)
        .with_target(false)
        .with_thread_ids(false)
        .with_thread_names(false)
        .with_file(false)
        .with_line_number(false)
        .with_ansi(true);

    // Initialize logging using RUST_LOG environment variable, defaulting to INFO level
    tracing_subscriber::fmt()
        .with_env_filter(EnvFilter::try_from_env("RUST_LOG").unwrap_or_else(|_| {
            EnvFilter::from_default_env().add_directive(tracing::Level::INFO.into())
        }))
        .event_format(format)
        .init();
}
```

proposer/succinct/Cargo.toml
```diff
@@ -23,6 +27,12 @@ op-succinct-host-utils.workspace = true
 # sp1
 sp1-sdk.workspace = true
 
+# alloy
+alloy-eips = { version = "0.11.1" }
+alloy-provider = { version = "0.11.1" }
+alloy-transport = { version = "0.11.1" }
+alloy-signer-local = { version = "0.11.1" }
```
fakedev9999, line 34: Let's use `.workspace = true.`

proposer/succinct/src/proposer.rs
```diff
@@ -0,0 +1,1013 @@
...
+            self.contract_config.l2oo_address,
+            self.driver_config.fetcher.as_ref(),
+        )
+        .await?;
+
+        // Get all completed range proofs from the database.
+        let completed_range_proofs = self
+            .driver_config
+            .driver_db_client
+            .fetch_completed_range_proofs(
```
ctian1, line 417: I think this query doesn't need to return proof bytes?

vcastellm (commented):
Any specific reason you're switching to Postgres?

PR #403, opened 2025-02-21
fix: various tweaks of the succinct fdg

smartcontracts:
Misc changes to the contracts that I would make if I had total control. Just a draft so that I can leave comments on the design decisions. Can be closed once you've decided what you like and don't like.


fakedev9999:
@smartcontracts All your changes makes sense! Will be closing this PR, but changes will be in #408 with fixes to tests to make it work + update to docs. Thanks again!

PR #405, opened 2025-02-22
fix(script): `upgrade-oracle` justfile

TimTinkers:
Without `env`, multiple environment variables separated by spaces would cause `L2OO_ADDRESS=0x...` to attempt to be executed as a command. This would cause the entire `upgrade-oracle` command to fail.


ratankaliani:
Thanks for the PR!

PR #406, opened 2025-02-22
fix: oracle upgrader can update from 0x0

TimTinkers:
If the implementation address is 0, the oracle upgrade script does not actually use the newly-deployed implementation.


ratankaliani (approved):
LGTM! Thanks for the PR

PR #410, opened 2025-02-26
feat: add Docker support for fault proof components

fakedev9999:
Support Docker for running the fault proof components with

- Docker Compose file to spin up both components' containers
- Dockerfile.proposer, Dockerfile.challenger with multi-stage build
- Documentation for running with Docker


ratankaliani (commented):
This PR should be structured as a Dockerfile for the proposer and one for the challenger and a Docker compose that runs both, rather than a shell script that optionally runs both.

Having a Dockerfile with a shell script entrypoint is a bit of an antipattern for this configuration, and ideally we should publish separate (independently versioned) images for proposer and challenger.

fault_proof/Dockerfile.challenger
```diff
@@ -0,0 +1,57 @@
...
+    build-essential \
+    git \
+    pkg-config \
+    libssl-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Rust
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH=/root/.cargo/bin:$PATH
+RUN rustup install nightly && rustup default nightly
```
ratankaliani, line 21: Don't believe nightly is necessary

ratankaliani (approved):
LGTM

PR #411, opened 2025-02-26
fix(fault_proof): fix defend logic

fakedev9999:
# Summary
This PR modifies the `get_oldest_game_address` function to accept a parameter that determines whether to check for equality or inequality between the output root and the claim.

# Problem
The `get_oldest_game_address` function is used for both defense (get_oldest_defensible_game_address) and challenging (get_oldest_challengable_game_address).

For challenging, we should challenge games where the claimed root doesn't match the actual output root (the current logic is correct for this)

For defending, we should defend games where the claimed root does match the actual output root (the current logic is incorrect for this)

So for defense, we want the opposite condition: `if output_root == game_claim`. This would mean we're defending games where the claim is correct but a challenge occurred.

### DIFF
--- a/fault_proof/src/lib.rs
+++ b/fault_proof/src/lib.rs
@@ -149,15 +149,17 @@ where
     async fn get_anchor_l2_block_number(&self, game_type: u32) -> Result<U256>;
 
     /// Get the oldest game address with a given condition.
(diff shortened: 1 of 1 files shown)


ratankaliani (approved):
LGTM!

PR #412, opened 2025-02-27
chore: unify solady

fakedev9999:
Unifies solady dependency to version v0.0.158, which is the version used by Optimism.

Necessary changes are made since LibBytes is not implemented in v0.0.158.


ratankaliani (approved):
LGTM

PR #413, opened 2025-02-28
feat(fault_proof): add intelligent retry mechanism to prove_game function

qvkare:
Added a smart retry mechanism to the `prove game` function to handle temporary failures during proof generation.

The implementation uses exponential backoff with 3 retry attempts and can distinguish between temporary errors (network issues, service unavailability) and permanent ones. This ensures our system doesn't lose games due to transient issues, making it more reliable in production.

The changes are minimal and focused on improving resilience without altering the core functionality.


qvkare:
Hey @ratankaliani, can I get your opinion on that?

PR #414, opened 2025-03-01
fix(fault_proof): fix env file from being overwritten

fakedev9999:
Fixes env file for proposer from being overwritten.

Env file is already loaded in bin/proposer.rs.

```rust
async fn main() {
    setup_logging();

    let args = Args::parse();
    dotenv::from_filename(args.env_file).ok();
```


ratankaliani (approved):
LGTM

PR #417, opened 2025-03-03
feat(proposer): add Agglayer integration

vcastellm:
We're introducing a new endpoint in the JSON RPC API, to support requesting agg proofs from external services. The purpose of this is to facilitate integration of external managers of proofs, in a way they can decide how to aggregate the more granular span proofs together.

Test the new method using:
cast rpc proofs_requestAggProof [start_block] [end_block] [l1_block_number] [l1_block_hash]


ratankaliani:
Closing for now, as we're not planning on merging in the short-term.

PR #420, opened 2025-03-04
feat: make misc changes to `OPSuccinctFaultDisputeGame`

fakedev9999:
Made misc changes to `OPSuccinctFaultDisputeGame`.

1. Removed redundant game resolution check in `closeGame` since the subsequent call to `ANCHOR_STATE_REGISTRY.isGameFinalized(IDisputeGame(address(this)))` verifies the game's resolution status.
2. Refactored to use = instead of += for bond distribution.
3. Added proposal status sanity check for resolution.
4. Added credit getter like OP's implementation.
5. Added dispute game factory check since there could be misalignment with the factory and the game impl contract.

### DIFF
--- a/contracts/src/fp/OPSuccinctFaultDisputeGame.sol
+++ b/contracts/src/fp/OPSuccinctFaultDisputeGame.sol
@@ -24,7 +24,6 @@ import {
     ClockTimeExceeded,
     GameNotFinalized,
     GameNotInProgress,
-    GameNotResolved,
     IncorrectBondAmount,
     InvalidBondDistributionMode,
     NoCreditToClaim,
@@ -507,12 +506,6 @@ contract OPSuccinctFaultDisputeGame is Clone, ISemver, IDisputeGame {
         } else if (bondDistributionMode != BondDistributionMode.UNDECIDED) {
             // We shouldn't get here, but sanity check just in case.
             revert InvalidBondDistributionMode();
-        }
-
-        // Make sure that the game is resolved.
-        // AnchorStateRegistry should be checking this but we're being defensive here.
-        if (resolvedAt.raw() == 0) {
-            revert GameNotResolved();
         }
 
         // Game must be finalized according to the AnchorStateRegistry.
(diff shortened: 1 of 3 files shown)


contracts/src/fp/OPSuccinctFaultDisputeGame.sol
```diff
@@ -597,6 +605,22 @@ contract OPSuccinctFaultDisputeGame is Clone, ISemver, IDisputeGame {
...
+    ////////////////////////////////////////////////////////////////
+
+    /// @notice Returns the credit balance of a given recipient.
+    /// @param _recipient The recipient of the credit.
+    /// @return credit_ The credit balance of the recipient.
+    function credit(address _recipient) external view returns (uint256 credit_) {
+        if (bondDistributionMode == BondDistributionMode.REFUND) {
+            credit_ = refundModeCredit[_recipient];
+        } else {
+            // Always return normal credit balance by default unless we're in refund mode.
```
ratankaliani, line 619:
```suggestion
            // Always return normal credit balance by default unless in refund mode.
```

ratankaliani (approved):
LGTM!

contracts/src/fp/OPSuccinctFaultDisputeGame.sol
```diff
@@ -434,23 +436,35 @@ contract OPSuccinctFaultDisputeGame is Clone, ISemver, IDisputeGame {
...
+                    normalModeCredit[claimData.prover] = address(this).balance;
+                }
+                // If the prover is different from the proposer, the proposer gets the initial bond back,
+                // and the prover gets the challenger's bond.
+                else {
+                    normalModeCredit[claimData.prover] = CHALLENGER_BOND;
+                    normalModeCredit[gameCreator()] = address(this).balance - CHALLENGER_BOND;
+                }
+            } else {
+                // We shouldn't get here, but sanity check just in case.
```
ratankaliani, line 466:
```suggestion
                // This edge case shouldn't be reached, sanity check just in case.
```

PR #431, opened 2025-03-11
feat: commit used memory from range program

leruaa:
This PR update the range program to commits the memory usage that can then be retrieved in the host for benchmark. This is used in the Base FP bench repo.

Also bump Kona version to include this fix: https://github.com/op-rs/kona/pull/1217


ratankaliani:
Closing because committing to the memory usage from within the program will cause issues with the current on-chain verification.

PR #432, opened 2025-03-11
wip: fix ci

fakedev9999:
Stats from running cost estimator is not being reported in CI. This PR fixes it.


fakedev9999:
Going to re-open to check if it works.

PR #433, opened 2025-03-11
ci: filter test properly

fakedev9999:
Stats from running execute_batch `scripts/prove/tests/multi.rs` is not being reported in CI.

Updated test code, since misconfiguration disabled some unit tests, failing to catch test error. See [failed ci job](https://github.com/succinctlabs/op-succinct/actions/runs/13796640953/job/38589741396).


ratankaliani (approved):
LGTM

PR #440, opened 2025-03-13
feat(validity): handle proof request timeout

fakedev9999:
If the proof requests are made asynchronously to the prover network, timeout should be handled by checking the deadline of the proof request.

### DIFF
--- a/validity/src/proposer.rs
+++ b/validity/src/proposer.rs
@@ -313,6 +313,18 @@ where
                 .get_proof_status(B256::from_slice(proof_request_id))
                 .await?;
 
+            // Check if current time exceeds deadline. If so, the proof has timed out.
+            let current_time = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_secs();
+            if current_time > status.deadline {
+                return Err(anyhow!(
+                    "Proof request has timed out for request id: {:?}",
+                    proof_request_id
+                ));
+            }
+
             let execution_status = ExecutionStatus::try_from(status.execution_status)
                 .context("Failed to convert execution status to ExecutionStatus.")?;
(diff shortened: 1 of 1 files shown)


validity/src/proposer.rs
```diff
@@ -311,6 +311,18 @@ where
...
+            let current_time = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_secs();
+            if current_time > status.deadline {
+                return Err(anyhow!(
+                    "Proof request has timed out for request id: {:?}",
+                    proof_request_id
+                ));
+            }
```
ratankaliani, line 324: If the proof times out, we should retry the request.

In the proposer/proof requester, unless we explicitly handle the status of the proof request, it's status will not be updated.

If you look at where `process_proof_request_status` is called, the Result is not processed (silent error).

This means that when you encounter a timeout, you should explicitly handle it.

This matches the logic we had here: https://github.com/succinctlabs/op-succinct/blob/0d51b56906240c5e3ba028622552650586761f00/proposer/succinct/bin/server.rs#L536-L554

ratankaliani (commented):
You need to explicitly retry the request if it has timed out.

validity/src/proposer.rs
```diff
@@ -311,13 +311,24 @@ where
...
+                .as_secs();
+            if current_time > status.deadline {
+                self.proof_requester
+                    .retry_request(request.clone(), status.execution_status())
+                    .await?;
+
+                return Err(anyhow!(
+                    "Proof request has timed out for request id: {:?}",
+                    proof_request_id
+                ));
```
ratankaliani, line 327: Don't need this, you can just return Ok(()).

The purpose of this function is to process the proof request status.

ratankaliani (approved):
LGTM!

PR #441, opened 2025-03-14
feat(validity): store `l1_head_block_number`

fakedev9999:
Would be nice to have `l1_head_block_number` stored alongside range proof requests to fully determine the source of the cost.


validity/src/db/client.rs
```diff
@@ -636,7 +638,9 @@ impl DriverDBClient {
                     .push_bind(req.total_tx_fees.clone())
                     .push_bind(req.l1_chain_id)
                     .push_bind(req.l2_chain_id)
-                    .push_bind(req.contract_address.as_ref().map(|arr| &arr[..]));
+                    .push_bind(req.contract_address.as_ref().map(|arr| &arr[..]))
+                    .push_bind(req.prover_address.as_ref().map(|arr| &arr[..]))
```
fakedev9999, line 642: Sneaked in a missing line here.

ratankaliani:
@fakedev9999 Yes

validity/migrations/01_init.sql
```diff
@@ -30,7 +30,8 @@ CREATE TABLE IF NOT EXISTS requests (
     l1_chain_id BIGINT NOT NULL,
     l2_chain_id BIGINT NOT NULL,
     contract_address BYTEA,
-    prover_address BYTEA
+    prover_address BYTEA,
+    l1_head_block_number BIGINT
```
ratankaliani, line 34: You should move these to a separate migration.

validity/src/db/types.rs
```diff
@@ -127,6 +128,11 @@ impl OPSuccinctRequest {
             .get_l2_block_data_range(start_block as u64, end_block as u64)
             .await?;
 
+        let l1_head_block_number = fetcher
+            .get_l1_head_with_safe_head(end_block as u64)
+            .await?
+            .1;
```
ratankaliani, line 134: If we're going to do it this way, then we should update the places where we get the host args to ensure that this actually matches.

Currently, we fetch the `l1_head` on the fly.

validity/src/db/client.rs
ratankaliani: Do you think there's a more type-safe way to do this to avoid this issue in the future?

fakedev9999: Hmm.. I don't think there is.

ratankaliani (approved):
LGTM!

PR #446, opened 2025-03-15
chore: fix migration

ratankaliani:


vcastellm:
@ratankaliani migration is failing after this change, extra comma at the end, aside from that, can you elaborate why you're removing tables?

Better to create a new migration instead of modifying existing ones.

PR #449, opened 2025-03-18
refactor: agg proof witgen

ratankaliani:


validity/src/proof_requester.rs
```diff
@@ -371,25 +381,19 @@ impl<H: OPSuccinctHost> OPSuccinctProofRequester<H> {
...
+                    request.end_block,
+                    B256::from_slice(request.checkpointed_l1_block_hash.as_ref().ok_or_else(
+                        || anyhow::anyhow!("Aggregation proof has no checkpointed block."),
+                    )?),
+                    request.l1_chain_id,
+                    request.l2_chain_id,
+                    Address::from_slice(request.prover_address.as_ref().ok_or_else(|| {
+                        anyhow::anyhow!("Prover address must be set for aggregation proofs.")
+                    })?),
+                )
```
fakedev9999, line 395: Can't we just input request here?

ratankaliani: The purpose of this PR is to expose a method for generating the witness that does not require a `Request` object, but rather the constituent parts.

Otherwise, the interface/abstraction is not super clear, as agg proof witnessgen is not tied to a `Request` object.

PR #450, opened 2025-03-18
feat: add contract upgrade script for fp

fakedev9999:
Adds just command to upgrade OPSuccinctFaultDisputeGame contract.

Has a `DRY_RUN` environment variable to allow simulation before actual upgrade.

### DIFF
--- a/contracts/script/fp/UpgradeOPSuccinctFDG.s.sol
+++ b/contracts/script/fp/UpgradeOPSuccinctFDG.s.sol
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+// Libraries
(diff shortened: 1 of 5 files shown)


contracts/script/fp/UpgradeOPSuccinctFDG.s.sol
ratankaliani, line 1: Can you add a method here which outputs the calldata for this? Most chains in production have a multi-sig or something similar and need the calldata to be outputted explicitly.

ratankaliani (approved):
LGTM! Include the calldata output in a separate PR.

contracts/script/fp/UpgradeOPSuccinctFDG.s.sol
fakedev9999: I'll do it in a separate PR!

PR #452, opened 2025-03-19
feat(proposer+contracts): validity `OptimismPortal2` integration

fakedev9999:
- Refactors `extraData` method in `OPSuccinctDisputeGame` to use `_getArgBytes` in [solady](https://github.com/Vectorized/solady/blob/502cc1ea718e6fa73b380635ee0868b0740595f0/src/utils/Clone.sol#L24-L40). 
- Updates `GameType` for validity game to use upstream.
- Add a lock to the DB to prevent initializing two proposers for the same DB.
- Fix the `extraData` encoding for the validity proposer.


ratankaliani (approved):
LGTM

PR #453, opened 2025-03-19
feat: add `handle_bond_claiming` for fp challenger

fakedev9999:
Now OP Succinct Lite challengers can claim bonds for the games that they have successfully challenged.

### DIFF
--- a/fault-proof/bin/challenger.rs
+++ b/fault-proof/bin/challenger.rs
@@ -35,6 +35,7 @@ where
     P: Provider<Ethereum> + Clone,
 {
     config: ChallengerConfig,
+    challenger_address: Address,
     l2_provider: L2Provider,
     l1_provider_with_wallet: L1ProviderWithWallet<F, P>,
     factory: DisputeGameFactoryInstance<(), L1ProviderWithWallet<F, P>>,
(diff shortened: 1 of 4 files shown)


fault-proof/bin/challenger.rs
```diff
@@ -122,6 +125,58 @@ where
             .await
     }
 
+    /// Handles claiming bonds from resolved games.
+    pub async fn handle_bond_claiming(&self) -> Result<Action> {
+        let _span = tracing::info_span!("[[Claiming Bonds]]").entered();
+
+        if let Some(game_address) = self
+            .factory
+            .get_oldest_claimable_bond_game_address(
```
ratankaliani, line 134: Where does this function exist on the `DisputeGameFactory`?

I don't see it here: https://github.com/succinctlabs/optimism/blob/develop/packages/contracts-bedrock/src/dispute/DisputeGameFactory.sol?

ratankaliani (approved):
LGTM

fault-proof/bin/challenger.rs
fakedev9999: It's an impl of a FactoryTrait! https://github.com/succinctlabs/op-succinct/blob/7db1acc6c5122dcdaabc8454a464ecea6c52de7d/fault-proof/src/lib.rs#L539-L575

PR #454, opened 2025-03-19
feat(fault-proof): better error gauges for fp

fakedev9999:
Instead of total error count gauges, there's an error gauge for each actions for proposer and challenger, so that it's easier to track down the source of issue.


ratankaliani (approved):
LGTM

PR #455, opened 2025-03-20
feat(contracts): output upgrade calldata on dry run

fakedev9999:
Now the upgrade script outputs upgrade calldata on dry run.

As mentioned in #450, most chains in production have a multi-sig or something similar and need the calldata to be outputted explicitly to upgrade.


ratankaliani (approved):
LGTM

PR #456, opened 2025-03-20
docs: improve fp docs

fakedev9999:
Improved fp docs after giving a final pass on a newly cloned op-succinct repo.


gatleas17 (changes requested):
pls help with changes

PR #459, opened 2025-03-22
ci: add cycle count diff

fakedev9999:
Adds cycle count diff CI job when the `RANGE_ELF_EMBEDDED` is updated.
This is to make sure we're not losing performance on every elf update.


ratankaliani (commented):
How is `base_cycle_stats.json` updated? Can you explain how this accurately shows the diff?

programs/range/src/main.rs
```diff
@@ -49,6 +49,7 @@ fn main() {
             .await
             .expect("failed to run client");
 
+        sp1_zkvm::io::commit(&BootInfoStruct::from(boot_info.clone()));
```
ratankaliani, line 52: The `range` program should not be modified?

ratankaliani (commented):
The range program shouldn't be modified.

programs/range/src/main.rs
fakedev9999: I intentionally made a change to the program to see if the CI works. Will drop the commit before merging.

ratankaliani (approved):
LGTM

PR #471, opened 2025-04-03
feat: gRPC server for remotely requesting agg proofs

vcastellm:
Still WIP Agg proofs gRPC service


ratankaliani:
Closing for now, as we're not planning on merging in the short-term.

PR #473, opened 2025-04-04
test: add test for OPSuccinctDisputeGame

fakedev9999:
Adds tests for OPSuccinctDistputeGame, which supports OptimismPortal2.
Including, test with a real proof data for Phala Testnet: [tx](https://sepolia.etherscan.io/tx/0xeb3ccf9d86b5495da24df4ecfbb02b03404ef3a72de4fc29326c996be4c10005).

Added usage of AccessManager for OPSuccinctDisputeGame for permission control. 

Since OPSuccinctDisputeGame is created through DisputeGameFactory with CWIA pattern and to make OPSuccinctDisputeGame work by wrapping over OPSuccinctL2OutputOracle, now the check for approved proposers inside OPSuccinctL2OutputOracle are done against tx.origin (not the msg.sender).


contracts/script/validity/OPSuccinctDGFDeployer.s.sol
```diff
@@ -11,17 +11,41 @@ import {console} from "forge-std/console.sol";
...
 
+        // Proposer must be permissionless or else the check in `proposeL2Output` will fail.
         l2OutputOracleProxy.addProposer(address(0));
 
+        // Deploy the access manager.
+        AccessManager accessManager = new AccessManager();
+        console.log("Access manager deployed: ", address(accessManager));
+        if (vm.envOr("PERMISSIONLESS_MODE", false)) {
+            accessManager.setProposer(address(0), true);
+            // TODO(fakedev9999): Figure out what's the case for challengers in OptimismPortal2 support.
```
ratankaliani, line 31: Can this be removed? There's no use for challengers in `OptimismPortal2`

contracts/script/validity/OPSuccinctDGFDeployer.s.sol
```diff
@@ -11,17 +11,41 @@ import {console} from "forge-std/console.sol";
...
+import {LibString} from "@solady/utils/LibString.sol";
 
 contract OPSuccinctDFGDeployer is Script, Utils {
     function run() public returns (address) {
         vm.startBroadcast();
 
         OPSuccinctL2OutputOracle l2OutputOracleProxy = OPSuccinctL2OutputOracle(vm.envAddress("L2OO_ADDRESS"));
 
+        // Proposer must be permissionless or else the check in `proposeL2Output` will fail.
         l2OutputOracleProxy.addProposer(address(0));
```
ratankaliani, line 24: This changes the security assumptions about the `l2OutputOracleProxy`, let's discuss more.

fault-proof/src/proposer.rs
```diff
@@ -131,7 +131,7 @@ where
             .compressed()
             .strategy(FulfillmentStrategy::Hosted)
             .skip_simulation(true)
-            .cycle_limit(1_000_000_000_000)
+            .cycle_limit(10_000_000_000)
```
fakedev9999, line 134: TODO: revert

contracts/src/validity/OPSuccinctL2OutputOracle.sol
```diff
@@ -309,7 +309,7 @@ contract OPSuccinctL2OutputOracle is Initializable, ISemver {
     /// @param _l2BlockNumber The L2 block number that resulted in _outputRoot.
     /// @param _l1BlockNumber The block number with the specified block hash.
     /// @param _proof The aggregation proof that proves the transition from the latest L2 output to the new L2 output.
-    /// @param _proverAddress The address of the prover that submitted the proof. Note: proverAddress is not required to be the msg.sender as there is no reason to front-run the prover.
+    /// @param _proverAddress The address of the prover that submitted the proof. Note: proverAddress is not required to be the tx.origin as there is no reason to front-run the prover.
```
ratankaliani, line 312: You should explain here under what conditions `tx.origin` is safe.

There are some attacks that can be made against `tx.origin`. As long as the relayers/proposers do not interact with untrusted contracts, `tx.origin` has equivalent security to `msg.sender`.

ratankaliani (commented):
Looks much cleaner! We should be specific on why `tx.origin` is safe.

contracts/src/validity/OPSuccinctL2OutputOracle.sol
```diff
@@ -381,7 +392,7 @@ contract OPSuccinctL2OutputOracle is Initializable, ISemver {
     {
         // The proposer must be explicitly approved, or the zero address must be approved (permissionless proposing).
         require(
-            approvedProposers[msg.sender] || approvedProposers[address(0)],
```
fakedev9999, line 384: @ratankaliani I'm thinking of keeping `msg.sender` for `whenOptimistic`. What's your thoughts here? I'm really not sure.

contracts/src/validity/OPSuccinctDisputeGame.sol
```diff
@@ -152,4 +152,14 @@ contract OPSuccinctDisputeGame is ISemver, Clone, IDisputeGame {
...
+
+    ////////////////////////////////////////////////////////////////
+    //                     IMMUTABLE GETTERS                      //
+    ////////////////////////////////////////////////////////////////
+
+    /// @notice Getter for the L2OutputOracle.
+    /// @return l2OutputOracle_ The address of the L2OutputOracle.
+    function l2OutputOracle() external view returns (address l2OutputOracle_) {
+        l2OutputOracle_ = L2_OUTPUT_ORACLE;
+    }
```
ratankaliani, line 164: Do we need this? Isn't `L2_OUTPUT_ORACLE` as an immutable address/variable already viewable?

contracts/src/validity/OPSuccinctL2OutputOracle.sol
ratankaliani: I'm fine to keep as is. You should add a note on `whenOptimistic` that it's not compatible with `OPSuccinctDisputeGame`.

contracts/src/validity/OPSuccinctDisputeGame.sol
fakedev9999: It's internal variable so the call to view will revert.

```solidity
address internal immutable L2_OUTPUT_ORACLE;
```

ratankaliani: Sg

ratankaliani (approved):
LGTM after you add the comment specifying that optimistic mode will not work with `OPSuccinctDisputeGame`.

PR #481, opened 2025-04-14
chore: improve validity quick-start devex

### LINKED ISSUE #480: add better tooling dependency management
Following https://succinctlabs.github.io/op-succinct/validity/quick-start.html on a fresh ec2 instance with ubuntu ami is quite painful. The dependency list is not complete, and there's a LOT of hidden dependencies. For example, `just deploy-oracle` actually requires having openssl and rocksdb (!) installed because of its dependency on kona-host:
```
$ cargo tree -i librocksdb-sys
librocksdb-sys v0.17.1+9.9.3
└── rocksdb v0.23.0
    └── kona-host v0.1.1 (https://github.com/op-rs/kona?tag=kona-client%2Fv0.1.0-beta.15#1d4c2fca)
        └── op-succinct-host-utils v0.1.0 (/home/ubuntu/op-succinct/utils/host)
            └── op-succinct-scripts v0.1.0 (/home/ubuntu/op-succinct/scripts/utils)
```

System libraries are hard to deal with, but would be good to at least in the short term describe them, perhaps add a `apt install -y ...` and `brew install -y ...` command, and in the long run try to get rid of the kona-host dependency for the utils that just need to get basic information (might require a refactor on kona-host's part though).

As for user space dependencies like forge, cast, etc., would be good to use nix or mise or whatever to track them and create a reproducible dev environment.

samlaf:
Partly Fixes #480 

Used mise.toml to pin the dependencies that you guys have in https://succinctlabs.github.io/op-succinct/validity/quick-start.html. Made it a separate commit such that it can be reverted if you guys don't want to rely on mise.toml. OP uses it everywhere and its pretty simple so I think its a nice to have. Plus its totally optional.

Also made some small improvements to the justfile commands such as automatically parsing the deployment files and populating the .env with the deployed contract addresses. Also some minor bash mistakes like not using quotes around some values which created errors when they were not part of .env file (like ETHERSCAN_API_KEY)

### DIFF
--- a/justfile
+++ b/justfile
@@ -81,7 +81,7 @@ deploy-mock-verifier env_file=".env":
     cd contracts
 
     VERIFY=""
-    if [ $ETHERSCAN_API_KEY != "" ]; then
+    if [ "$ETHERSCAN_API_KEY" != "" ]; then
       VERIFY="--verify --verifier etherscan --etherscan-api-key $ETHERSCAN_API_KEY"
     fi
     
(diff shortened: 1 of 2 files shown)


ratankaliani:
Thanks for the PR @samlaf! Appreciate you going through the quickstart and pointing out the issues on a fresh instance, I've incorporated the changes into https://github.com/succinctlabs/op-succinct/pull/485.

PR #483, opened 2025-04-14
chore: bump hana

fakedev9999:
Bumps hana from v0.1.0-beta.6 to v0.1.0-beta.7.

There was a patch in hana to fix `end share x is higher than block shares 256` error while fetching share proof.


ratankaliani (approved):
LGTM

PR #487, opened 2025-04-17
feat(utils): handle eth da fallback

fakedev9999:
Adds branching logic to handle ETH DA fallback.

Version byte will be 0 if ETH DA batch and will be 1 if Alt DA batch.

See https://specs.optimism.io/protocol/derivation.html#batcher-transaction-format for more details.

### DIFF
--- a/utils/host/src/hosts/celestia.rs
+++ b/utils/host/src/hosts/celestia.rs
@@ -137,20 +137,33 @@ impl OPSuccinctHost for CelestiaOPSuccinctHost {
                     if to_addr == batch_inbox_address {
                         let calldata = tx.input();
 
-                        // Check that the DA layer byte prefix is correct.
-                        // https://github.com/ethereum-optimism/specs/discussions/135.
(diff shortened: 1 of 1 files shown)


utils/host/src/hosts/celestia.rs
```diff
@@ -137,20 +137,33 @@ impl OPSuccinctHost for CelestiaOPSuccinctHost {
...
+                            let height_bytes = &calldata[3..11];
+                            let celestia_height =
+                                u64::from_le_bytes(height_bytes.try_into().unwrap());
+
+                            if celestia_height < latest_celestia_block {
+                                found_valid_tx = true;
+                                l2_block_number = Some(l2_safe_head_number);
+                            }
+                        } else {
+                            return Err(anyhow!("Invalid version byte for batcher transaction"));
```
ratankaliani, line 166: Can you split this into a separate function that returns the height given the version byte?

ratankaliani (approved):
LGTM

PR #488, opened 2025-04-17
ci: fix cycle count diff ci

fakedev9999:
Fixes cycle count diff ci. Tested in #491.

We used to store base_cycle_stats.json and had the bot to update the file, push the diff to the file, but now we don't have a file tracked anymore instead run both on old and new branch and post the diff result to github.


ratankaliani (approved):
LGTM

PR #498, opened 2025-04-23
refactor: move zkvm ELFs into a separate crate

iljakuklic:
This makes it easier for third party software to use the zkvm ELF binary data (e.g. for the purpose of calculating the verification keys) without having to involve the heavy dependencies of `op-succinct-host-utils`.


utils/elfs/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+//! The zkvm ELF binaries.
```
ratankaliani, line 1:
```suggestion
//! The ELF binaries.
```

fakedev9999:
Thanks! Closing in favor of #508. Cherry-picked and rebased.

PR #503, opened 2025-05-02
docs: guide with env var value

fakedev9999:
Updating docs with actual env var value to set for configuration will make configuration more straightforward.


ratankaliani (approved):
LGTM

PR #514, opened 2025-05-20
feat: add web3 signer support for op-succinct-lite

fakedev9999:
Adds web3 signer support for op-succinct-lite proposer and challenger.

- [x] Test proposer with private key signer: https://sepolia.etherscan.io/tx/0xcb1494239a58afbe6191d080b3497578688ec699a27d78e88cbb9b64021619fe
- [x] Test proposer with web3 signer: https://sepolia.etherscan.io/tx/0x09ebfe82460f82de4e1eeee28781cd39740e47743b37e01270a5b83827563f50
- [x] Test challenger with private key signer: https://sepolia.etherscan.io/tx/0xd5ba1f67751904784b233bd5adf45f5ff51a1fd2cf38ced8072361062f654328
- [x] Test challenger with web3 signer: https://sepolia.etherscan.io/tx/0x0d7b14c512f91c6a86a168c8648c5cd6ea0cac0699396b54789390844299d4fc

### DIFF
--- a/book/fault_proofs/quick_start.md
+++ b/book/fault_proofs/quick_start.md
@@ -31,8 +31,12 @@ This guide provides the fastest path to try out OP Succinct fault dispute games 
     DISPUTE_GAME_FINALITY_DELAY_SECONDS=604800
     MAX_CHALLENGE_DURATION=604800
     MAX_PROVE_DURATION=86400
-    STARTING_ROOT=0xbd4ab027af3c4db0c6b1329dad6d9f8e2505accabec75bfaa2b6b8033c1e60c5...
+    STARTING_ROOT=0xbd4ab027af3c4db0c6b1329dad6d9f8e2505accabec75bfaa2b6b8033c1e60c5
     STARTING_L2_BLOCK_NUMBER=791000
+
+    # Optional
+    # For production setting, set PROPOSER_ADDRESSES and CHALLENGER_ADDRESSES properly.
+    PERMISSIONLESS_MODE=true
 
     # For testing, use mock verifier
     USE_SP1_MOCK_VERIFIER=true
@@ -54,7 +58,9 @@ Save the output addresses, particularly the `FACTORY_ADDRESS` output as "Factory
     ```env
     # Required Configuration
     L1_RPC=<YOUR_L1_RPC_URL>
+    L1_BEACON_RPC=<L1_BEACON_RPC_URL>
     L2_RPC=<YOUR_L2_RPC_URL>
+    L2_NODE_RPC=<L2_NODE_RPC_URL>
     FACTORY_ADDRESS=<FACTORY_ADDRESS_FROM_DEPLOYMENT>
     GAME_TYPE=42
     PRIVATE_KEY=<YOUR_PRIVATE_KEY>
@@ -72,8 +78,6 @@ Save the output addresses, particularly the `FACTORY_ADDRESS` output as "Factory
     ```env
     FAST_FINALITY_MODE=true
     NETWORK_PRIVATE_KEY=0x...
-    L1_BEACON_RPC=<L1_BEACON_RPC_URL>
-    L2_NODE_RPC=<L2_NODE_RPC_URL>
     ```
 
 To get a whitelisted key on the Succinct Prover Network for OP Succinct, fill out this [form](https://docs.google.com/forms/d/e/1FAIpQLSd-X9uH7G0bvXH_kjptnQtNil8L4dumrVPpFE4t8Ci1XT1GaQ/viewform). The Succinct team will reach out to you with an RPC endpoint you can use.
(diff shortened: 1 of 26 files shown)


book/fault_proofs/quick_start.md
```diff
@@ -31,9 +31,13 @@ This guide provides the fastest path to try out OP Succinct fault dispute games
     DISPUTE_GAME_FINALITY_DELAY_SECONDS=604800
     MAX_CHALLENGE_DURATION=604800
     MAX_PROVE_DURATION=86400
-    STARTING_ROOT=0xbd4ab027af3c4db0c6b1329dad6d9f8e2505accabec75bfaa2b6b8033c1e60c5...
+    STARTING_ROOT=0xbd4ab027af3c4db0c6b1329dad6d9f8e2505accabec75bfaa2b6b8033c1e60c5
     STARTING_L2_BLOCK_NUMBER=791000
 
+    # Optional
+    # For production setting, set PROPOSER_ADDRESSES and CHALLENGER_ADDRESSES properly.
```
ctian1, line 38: nit: could use more clarification ex. "Warning: setting this to true allows anyone to _. In a production setting, we recommend setting this to false and configuring PROPOSER_ADDRESSES and CHALLENGER_ADDRESSES accordingly."

utils/signer/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+                .with_timeout(Some(Duration::from_secs(TIMEOUT_SECONDS)))
+                .get_receipt()
+                .await?;
+
+            Ok(receipt)
+        }
+    }
+}
+
+/// A wrapper around `TransactionRequest` that adds a data field as bytes.
```
ctian1, line 124: This comment is a bit confusing. Is it just that TransactionRequest doesn't have a data field, but web3signer expects it? That makes sense, but it doesn't seem like a bug in alloy? It's just not what TransactionRequest is for

utils/signer/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+/// 2. Web3Signer's `eth_signTransaction` method specifically expects a "data" field in the JSON-RPC
+///    request, not the "input" field.
+/// 3. While the alloy-rpc-types-eth crate has methods like `normalize_data()` and `set_both()` to
+///    work with both fields, these are only implemented on the specific Ethereum transaction type
+///    struct and aren't accessible through the generic `N::TransactionRequest` interface we're
+///    using here.
+///
+/// TODO(fakedev9999): Once alloy fixes this, we can remove this wrapper.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TransactionRequestWrapper {
```
ctian1, line 138: Is this still necessary since it's not generic on Network anymore, and TransactionRequest.input has both data and input?

utils/signer/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+        } else {
+            anyhow::bail!(
+                "Neither (SIGNER_URL and SIGNER_ADDRESS) nor PRIVATE_KEY are set in environment"
+            )
+        }
+    }
+}
+
+/// Sign a transaction request using the configured `signer`.
+pub async fn sign_transaction_request_inner(
```
ctian1, line 56: This should be in `impl Signer` and take in &self. Also, name/comment are a bit misleading: it says sign but it also sends and awaits the transaction. Maybe it should just be called `send_transaction_request`?

utils/signer/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
```
ctian1, line 1: nit: IMO it might be a bit too granular to have a crate just for signer utils, given that validity and fault-proof are the only consumers. Maybe this stuff could just live in utils/host.

fakedev9999: I agree that it's a bit granular, but IMO utils/host is not a proper place for signer code to live in.

In the long term, I think there's a lot of room to refactor the code with shared utils such as signer across validity proposer and fault-proof proposer&challenger.

So I'm thinking of keeping as is, but in the future maybe renaming the crate to something like op-succinct-core-utils.

fault-proof/bin/challenger.rs
```diff
@@ -84,7 +81,7 @@ where
         let transaction_request =
             game.challenge().value(self.challenger_bond).into_transaction_request();
 
-        let receipt = self.sign_transaction_request(transaction_request).await?;
+        let receipt = self.send_transaction_request(transaction_request).await?;
```
fakedev9999, line 84:
```suggestion
self.signer
            .send_transaction_request_(self.config.l1_rpc.clone(), transaction_request)
            .await
```

ctian1 (approved):

fault-proof/bin/challenger.rs
```diff
@@ -64,17 +64,14 @@ where
...
     ) -> Result<TransactionReceipt> {
-        sign_transaction_request_inner(
-            self.signer.clone(),
-            self.config.l1_rpc.clone(),
-            transaction_request,
-        )
-        .await
+        self.signer
+            .send_transaction_request_inner(self.config.l1_rpc.clone(), transaction_request)
+            .await
```
fakedev9999, line 74: remove.

PR #540, opened 2025-06-23
fix: refactor fault-proof proposer to use task-based concurrency

fakedev9999:
- Convert sequential operations to concurrent tasks using task tracking
- Fix dashboard showing 0 values during proof generation
- Implement task spawning for game creation, defense, resolution, and bond claiming
- Add centralized task management with completion handling
- Ensure metrics collection runs independently of other operations

This prevents long-running operations (like witness generation for proof generation) from blocking metrics updates and other proposer activities.

### DIFF
--- a/fault-proof/bin/proposer.rs
+++ b/fault-proof/bin/proposer.rs
@@ -66,7 +66,7 @@ async fn main() -> Result<()> {
     // Initialize the metrics gauges.
     ProposerGauge::init_all();
 
-    proposer.clone().run().await.expect("Runs in an infinite loop");
+    proposer.run().await.expect("Runs in an infinite loop");
 
     Ok(())
 }
--- a/fault-proof/bin/proposer.rs
+++ b/fault-proof/bin/proposer.rs
@@ -51,10 +51,11 @@ async fn main() -> Result<()> {
 
     let fetcher = OPSuccinctDataFetcher::new_with_rollup_config().await?;
     let host = initialize_host(Arc::new(fetcher.clone()));
-    let proposer =
+    let proposer = Arc::new(
         OPSuccinctProposer::new(prover_address, proposer_signer, factory, Arc::new(fetcher), host)
             .await
-            .unwrap();
+            .unwrap(),
+    );
 
     // Initialize proposer gauges.
     ProposerGauge::register_all();
@@ -65,7 +66,7 @@ async fn main() -> Result<()> {
     // Initialize the metrics gauges.
     ProposerGauge::init_all();
 
-    proposer.run().await.expect("Runs in an infinite loop");
+    proposer.clone().run().await.expect("Runs in an infinite loop");
 
     Ok(())
 }
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -39,8 +39,30 @@ use crate::{
 /// Type alias for task ID
 pub type TaskId = u64;
 
+/// Wrapper for different types of task handles
+pub enum TaskHandle {
+    Async(tokio::task::JoinHandle<Result<()>>),
+    Blocking(tokio::task::JoinHandle<Result<()>>),
+}
+
+impl TaskHandle {
+    pub fn is_finished(&self) -> bool {
+        match self {
+            TaskHandle::Async(handle) => handle.is_finished(),
+            TaskHandle::Blocking(handle) => handle.is_finished(),
+        }
+    }
+
+    pub async fn await_result(self) -> std::result::Result<Result<()>, tokio::task::JoinError> {
+        match self {
+            TaskHandle::Async(handle) => handle.await,
+            TaskHandle::Blocking(handle) => handle.await,
+        }
+    }
+}
+
 /// Type alias for a map of task IDs to their join handles and associated task info
-pub type TaskMap = HashMap<TaskId, (tokio::task::JoinHandle<Result<()>>, TaskInfo)>;
+pub type TaskMap = HashMap<TaskId, (TaskHandle, TaskInfo)>;
 
 /// Information about a running task
 #[derive(Clone, Debug)]
@@ -561,7 +583,7 @@ where
         // Process completed tasks
         for id in completed {
             if let Some((handle, info)) = tasks.remove(&id) {
-                match handle.await {
+                match handle.await_result().await {
                     Ok(Ok(())) => {
                         tracing::debug!("Task {:?} completed successfully", info);
                     }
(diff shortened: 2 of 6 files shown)


fault-proof/bin/proposer.rs
```diff
@@ -65,7 +66,7 @@ async fn main() -> Result<()> {
     // Initialize the metrics gauges.
     ProposerGauge::init_all();
 
-    proposer.run().await.expect("Runs in an infinite loop");
+    proposer.clone().run().await.expect("Runs in an infinite loop");
```
yuwen01: is this needed? proposer is only used once right

fault-proof/src/proposer.rs
```diff
@@ -480,53 +506,330 @@ where
...
+        }
+        Ok(())
+    }
+
+    /// Spawn pending operations if not already running
+    async fn spawn_pending_operations(&self) -> Result<()> {
+        // Check if we should create a game and spawn task if needed
+        if !self.has_active_task_of_type(&TaskInfo::GameCreation { block_number: U256::ZERO }).await
+        {
+            if let Err(e) = self.spawn_game_creation_task().await {
```
yuwen01: seems like errors will be extremely common, if there are no games to check. can we only return an error in case of an actual error instead?

fault-proof/src/proposer.rs
```diff
@@ -480,53 +506,330 @@ where
...
+        }
+    }
+
+    /// Spawn a game creation task if conditions are met
+    #[tracing::instrument(name = "[[Proposing]]", skip(self))]
+    async fn spawn_game_creation_task(&self) -> Result<()> {
+        // First check if we should create a game
+        let should_create = self.should_create_game().await?;
+        if !should_create {
+            return Err(anyhow::anyhow!("No game creation needed"));
```
yuwen01: for example i think this shouldnt be an error

yuwen01: maybe do like result<taskcreationstatus> or something?

fault-proof/src/proposer.rs
```diff
@@ -480,53 +506,330 @@ where
...
+            matches!(info, TaskInfo::GameProving { game_address: addr } if *addr == game_address)
+        })
+    }
+
+    /// Spawn a game proving task for a specific game
+    async fn spawn_game_proving_task(&self, game_address: Address) -> Result<()> {
+        let proposer: OPSuccinctProposer<P, H> = self.clone();
+        let task_id = self.next_task_id.fetch_add(1, Ordering::Relaxed);
+
+        let handle = tokio::spawn(async move {
```
yuwen01: hmmmm this should be spawn_blocking for mock mode, since we're executing which is expensive and needs to happen to make progress, but spawn is fine for the real network prover.

maybe if we don't make too many, we can just do spawn blocking?

utils/celestia/host/src/blobstream_utils.rs
```diff
@@ -306,14 +306,14 @@ pub async fn get_highest_finalized_l2_block(
             }
             Err(e) => {
                 // Error occurred (e.g., indexer error), treat as unavailable and search lower
-                tracing::warn!("Error checking L2 block {}: {}, treating as unavailable", mid, e);
```
yuwen01: can we keep this a warn

utils/celestia/host/src/blobstream_utils.rs
```diff
@@ -128,23 +128,23 @@ async fn find_minimum_blobstream_block(
...
-                        tracing::info!(
+                        tracing::debug!(
                             "DataCommitmentStored event at L1 block {} does not contain Celestia height {}, continuing search",
                             block_number,
                             celestia_height
                         );
                         // Continue searching for the correct event
                     }
                     Err(e) => {
-                        tracing::warn!(
```
yuwen01: keep warn?

fault-proof/src/proposer.rs
```diff
@@ -480,53 +491,401 @@ where
...
+    /// Check if there's an active task of the given type
+    async fn has_active_task_of_type(&self, task_type: &TaskInfo) -> bool {
+        let tasks = self.tasks.lock().await;
+        tasks
+            .values()
+            .any(|(_, info)| std::mem::discriminant(info) == std::mem::discriminant(task_type))
+    }
+
+    /// Log current task statistics
+    async fn log_task_stats(&self) {
```
yuwen01, line 625: in prod, the interval is pretty long, so i think we should also log the task types being worked on.

fault-proof/src/proposer.rs
```diff
@@ -480,53 +491,401 @@ where
...
+                    ProposerGauge::GamesCreated.increment(1.0);
+                    Ok(())
+                }
+                Ok(None) => Ok(()),
+                Err(e) => Err(e),
+            }
+        }));
+
+        // Get the next proposal block for task info
+        let next_block = self.get_next_proposal_block().await.unwrap_or(U256::ZERO);
```
yuwen01, line 679: can we add this to logs

fault-proof/src/proposer.rs
```diff
@@ -28,16 +36,56 @@ use crate::{
     Action, FactoryTrait, L1Provider, L2Provider, L2ProviderTrait, Mode,
 };
 
+/// Type alias for task ID
+pub type TaskId = u64;
+
+/// Wrapper for different types of task handles
+pub enum TaskHandle {
+    Async(tokio::task::JoinHandle<Result<()>>),
```
yuwen01: if these are the same type then why do we need two?

fault-proof/src/proposer.rs
```diff
@@ -480,53 +491,401 @@ where
...
+            matches!(info, TaskInfo::GameProving { game_address: addr } if *addr == game_address)
+        })
+    }
+
+    /// Spawn a game proving task for a specific game
+    async fn spawn_game_proving_task(&self, game_address: Address) -> Result<()> {
+        let proposer: OPSuccinctProposer<P, H> = self.clone();
+        let task_id = self.next_task_id.fetch_add(1, Ordering::Relaxed);
+
+        // In mock mode, use spawn_blocking for CPU-intensive proof generation
```
yuwen01: nit: execution, or "mock proof generation"

fault-proof/src/proposer.rs
```diff
@@ -480,53 +491,401 @@ where
...
+    /// Check if there's an active proving task for a specific game
+    async fn has_active_proving_for_game(&self, game_address: Address) -> bool {
+        let tasks = self.tasks.lock().await;
+        tasks.values().any(|(_, info)| {
+            matches!(info, TaskInfo::GameProving { game_address: addr } if *addr == game_address)
+        })
+    }
+
+    /// Spawn a game proving task for a specific game
+    async fn spawn_game_proving_task(&self, game_address: Address) -> Result<()> {
```
yuwen01, line 780: is there any way to associate game addresses with block ranges? it's hard to get a sense if there's any progress being made / what ranges to retry

yuwen01 (changes requested):

PR #541, opened 2025-06-24
fix: record proposal permissioning bug

yuwen01:


contracts/test/fp/OPSuccinctFaultDisputeGame.t.sol
```diff
@@ -706,4 +707,72 @@ contract OPSuccinctFaultDisputeGameTest is Test {
...
+        vm.prank(unauthorizedUser);
+        vm.deal(unauthorizedUser, 1 ether);
+        factory.create{value: 1 ether}(
+            gameType, Claim.wrap(keccak256("new-claim-3")), abi.encodePacked(uint256(4000), uint32(1))
+        );
+
+        // After the new game, timeout resets - unauthorized user should not be allowed immediately
+        vm.prank(unauthorizedUser);
+        vm.deal(unauthorizedUser, 1 ether);
+        // vm.expectRevert(BadAuth.selector);
```
fakedev9999: Shouldn't this be uncommented?

contracts/test/fp/OPSuccinctFaultDisputeGame.t.sol
```diff
@@ -706,4 +707,72 @@ contract OPSuccinctFaultDisputeGameTest is Test {
...
```
fakedev9999: Shouldn't this be uncommented?

contracts/src/fp/AccessManager.sol
```diff
@@ -86,15 +100,16 @@ contract AccessManager is Ownable {
...
-    function recordProposal() external {
-        lastProposalTimestamp = block.timestamp;
-    }
-
     /// @notice Returns whether proposal fallback timeout has elapsed.
     /// @return Whether permissionless proposing is active.
     function isProposalPermissionlessMode() external view returns (bool) {
+        // Get the latest game to check its timestamp.
+        GameType gameType = GameType.wrap(42);
+        IDisputeGameFactory.GameSearchResult[] memory games = DISPUTE_GAME_FACTORY.findLatestGames(gameType, 0, 1);
```
fakedev9999: Are you sure this is tested? I think this is misimplementation?

contracts/src/fp/AccessManager.sol
```diff
@@ -68,12 +76,33 @@ contract AccessManager is Ownable {
...
+    function getLastProposalTimestamp() public view returns (uint256) {
+        // Get the latest game to check its timestamp.
+        GameType gameType = GameType.wrap(42);
+        uint256 numGames = DISPUTE_GAME_FACTORY.gameCount();
+        if (numGames == 0) {
+            return DEPLOYMENT_TIMESTAMP;
+        } else {
+            // Find the most recent game of the given type.
+            IDisputeGameFactory.GameSearchResult[] memory games =
+                DISPUTE_GAME_FACTORY.findLatestGames(gameType, numGames - 1, 1);
```
fakedev9999: What's your thoughts of having our own impl instead of using findLatestGames for gas optimization? Dgf's impl is not designed for finding a single game like our case so I see a lot of gas wasted in their impl.

fakedev9999 (approved):
LGTM

PR #549, opened 2025-07-02
test: improve e2e testing with anvil

fakedev9999:
Added comprehensive end-to-end tests for fault-proof system that run actual proposer and challenger binaries.

Key Features:
- Full Lifecycle Coverage: Tests complete game lifecycle from creation → challenge → resolution → bond claims
- Anvil Fork: Uses L1 fork with time manipulation to compress days of waiting into seconds
- Two Main Test Scenarios:
  a. Honest Proposer: Creates valid games, resolves unchallenged games, claims bonds
  b. Honest Challenger: Challenges invalid games, wins disputes, claims bonds

Implementation details:
- Contract Bindings: Auto-generated Forge bindings for contract interaction
- Time Warping: Strategic time manipulation to test different phases
- Event Monitoring: Tracks game state transitions and verifies expected outcomes

Fixes made:
- Fixed off-by-one error in bond claiming logic that prevented claiming from all games
- Fetch timestamp from l1 provider.


fault-proof/src/lib.rs
```diff
@@ -461,9 +463,10 @@ where
                 continue;
             }
 
-            let current_timestamp = l2_provider
-                .get_l2_block_by_number(BlockNumberOrTag::Latest)
+            let current_timestamp = l1_provider
```
yuwen01, line 465: why is this changed?

book/fault_proofs/testing.md
```diff
@@ -1,61 +1,173 @@
...
+
+### Test Configuration
+
+Key test constants defined in `fault_proof/tests/common/contracts.rs`:
+- **TEST_GAME_TYPE**: 42 (test-specific game type)
+- **INIT_BOND**: 0.01 ETH (minimal bond for testing)
+- **CHALLENGER_BOND**: 1 ETH (bond required for challenges)
+- **MAX_CHALLENGE_DURATION**: 1 hour (time to submit a challenge)
+- **MAX_PROVE_DURATION**: 12 hours (time to submit proof after challenge)
+- **AIRGAP**: 7 days (delay before bonds can be claimed)
```
yuwen01: can we make this DISPUTE_GAME_FINALITY to make it consistent

fault-proof/tests/common/anvil.rs
```diff
@@ -0,0 +1,113 @@
...
+    let (_, l1_block_number) = fetcher
+        .get_safe_l1_block_for_l2_block(target_l2)
+        .await
+        .context("Failed to get safe L1 block")?;
+
+    Ok(l1_block_number)
+}
+
+/// Cleanup the Anvil instance
+pub fn cleanup_anvil() {
```
yuwen01: why do we need to do this?

yuwen01: shouldnt it just get dropped and cleaned up when we exit the program?

fault-proof/tests/common/contracts.rs
```diff
@@ -0,0 +1,263 @@
...
+    // Deploy AnchorStateRegistry implementation
+    debug!("Deploying AnchorStateRegistry implementation...");
+    let anchor_impl = AnchorStateRegistry::deploy(provider.clone()).await?;
+    let anchor_impl_addr = *anchor_impl.address();
+    info!("✓ AnchorStateRegistry implementation deployed at: {}", anchor_impl_addr);
+
+    // Prepare initialization data with expected test values
+    // NOTE: We use the finalized L2 block number - 100 as the starting anchor root and test 3 games
+    // with proposal interval of 10 blocks.
+    let l2_block_number = U256::from(
```
yuwen01, line 77: same comment as before, i think op-succinct-infra can be used to detect issues with a live chain, and the e2e test can be used to test the core logic of the proposer / challenger

fault-proof/tests/common/contracts.rs
```diff
@@ -0,0 +1,263 @@
...
+    // Deploy AnchorStateRegistry proxy
+    debug!("Deploying AnchorStateRegistry proxy...");
+    let anchor_proxy =
+        ERC1967Proxy::deploy(provider.clone(), anchor_impl_addr, init_data.into()).await?;
+    let anchor_state_registry = *anchor_proxy.address();
+    info!("✓ AnchorStateRegistry proxy deployed at: {}", anchor_state_registry);
+
+    // 4. Deploy AccessManager
+    debug!("Deploying AccessManager...");
+    // Note: AccessManager deployment may fail on Anvil forks, so we handle it gracefully
```
yuwen01: why does this fail and anchor state registry not fail?

fault-proof/tests/common/contracts.rs
```diff
@@ -0,0 +1,263 @@
...
+
+/// Configure contracts after deployment
+pub async fn configure_contracts(
+    provider: L1Provider,
+    contracts: &DeployedContracts,
+) -> Result<()> {
+    info!("Configuring contracts for E2E testing");
+
+    // Only configure if all contracts were deployed
+    if contracts.access_manager == Address::ZERO {
```
yuwen01: why does this only check the access manager

fault-proof/tests/common/process.rs
```diff
@@ -0,0 +1,299 @@
...
+/// Configuration for starting a binary process
+pub struct ProcessConfig {
+    pub name: String,
+    pub binary_path: PathBuf,
+    pub env_vars: HashMap<String, String>,
+    pub log_stdout: bool,
+}
+
+/// Start a proposer binary with the given configuration
+pub async fn start_proposer_binary(
```
yuwen01: can we make a proposer and call proposer.run() instead? starting a binary like this will make it hard to catch errors and make it easier to gracefully stop the binary when the process exists.

yuwen01: might need to make a OPSuccinctProposer::test() -> OPSuccinctProposer method or something

fault-proof/tests/common/monitor.rs
```diff
@@ -0,0 +1,364 @@
...
+/// Wait for games to be resolved
+pub async fn wait_for_resolutions<P: Provider>(
+    provider: &P,
+    tracked_games: &[TrackedGame],
+    timeout_duration: Duration,
+) -> Result<Vec<u8>> {
+    info!("Waiting for {} games to be resolved...", tracked_games.len());
+
+    let deadline = tokio::time::Instant::now() + timeout_duration;
+    let mut statuses = vec![0u8; tracked_games.len()]; // 0 = IN_PROGRESS
```
yuwen01: can we do like Status::IN_PROGRES::into() or osmething instead?

fault-proof/tests/common/monitor.rs
```diff
@@ -0,0 +1,364 @@
...
+        if all_claimed {
+            return Ok(claims);
+        }
+
+        sleep(Duration::from_secs(2)).await;
+    }
+}
+
+/// Game status constants
+pub const GAME_STATUS_IN_PROGRESS: u8 = 0;
```
yuwen01: is it possible to rug this from the bindings instead of redefining here?

fault-proof/tests/common/monitor.rs
```diff
@@ -0,0 +1,364 @@
...
+
+            // If both credits are zero, the claim has been made
+            if normal_credit == U256::ZERO && refund_credit == U256::ZERO {
+                claims[i] = true;
+                info!("Bonds claimed for game {} (both credit balances are 0)", game.address);
+            } else {
+                all_claimed = false;
+
+                // Log current credit balances for debugging
+                if i == 0 && tokio::time::Instant::now().elapsed().as_secs().is_multiple_of(10) {
```
yuwen01: isn't tokio::time::Instant::now().elapsed() always zero? 

https://docs.rs/tokio/latest/tokio/time/struct.Instant.html#method.elapsed

fault-proof/tests/common/monitor.rs
```diff
@@ -0,0 +1,364 @@
...
+    Ok(())
+}
+
+/// Verify all games resolved correctly (proposer wins)
+pub fn verify_all_resolved_correctly(statuses: &[u8]) -> Result<()> {
+    verify_games_resolved(statuses, GAME_STATUS_DEFENDER_WINS, "ProposerWins")
+}
+
+/// Verify all bonds were claimed
+pub fn verify_all_bonds_claimed(claims: &[bool]) -> Result<()> {
```
yuwen01:
```pub fn verify_all_bonds_claimed(claims: &[bool]) -> Result<()> {
    if let Some((i, _)) = claims.iter().enumerate().find(|(_, &claimed)| !claimed) {
        anyhow::bail!("Game {} bonds were not claimed", i);
    }
    info!("All {} games had bonds claimed", claims.len());
    Ok(())
}```

File: fault-proof/tests/common/process.rs [hunk 12]
```diff
@@ -0,0 +1,299 @@
...
+    // Test-specific configuration
+    env.insert("FETCH_INTERVAL".to_string(), "2".to_string()); // Check more frequently in tests
+    env.insert("MAX_GAMES_TO_CHECK_FOR_CHALLENGE".to_string(), "10".to_string()); // Check more games
+
+    env
+}
+
+/// Wait for a process to produce a specific log line
+#[allow(dead_code)]
+pub async fn wait_for_log_line(
```
yuwen01: unused, this also looks like it would be hard to use anyway. Imagine changing the logging a little bit and breaking your test 😬

fault-proof/tests/e2e.rs
```diff
@@ -1,245 +1,289 @@
...
+    let mut challenger = start_challenger_binary(challenger_binary, challenger_env).await?;
+    info!("✓ Challenger service started");
 
-    // Spawn the challenger process first
-    tracing::info!("Spawning challenger");
-    let mut challenger_process = TokioCommand::new("cargo")
-        .args(["run", "--bin", "challenger"])
-        .spawn()
-        .expect("Failed to spawn challenger");
+    // === PHASE 1: Create Invalid Games ===
```
yuwen01, line 145: i thought we couldnt do invalid games yet

fault-proof/src/lib.rs
fakedev9999: @yuwen01 It's basically 2 reasons:
1. The contracts live on L1, so it makes more sense to fetch timestamp from L1.
2. The test will timeout(or take a long long time) and since we're warping time on forked L1.

fault-proof/tests/common/anvil.rs
fakedev9999: @yuwen01 I just copy pasted from the repo. Will double check.

fault-proof/tests/e2e.rs
fakedev9999: @yuwen01 We can't do malicious challenges, which requires defend with proof, which requires debug methods not supported by anvil.

fault-proof/tests/common/anvil.rs
fakedev9999: @yuwen01 The AnvilInstance already implements Drop trait, which will automatically clean up the AnvilInstance when the program exits. Will remove.

fault-proof/tests/common/process.rs
fakedev9999: @yuwen01 I think the current implementation already captures errors effectively:
- Line 151-163: We check if the process exits immediately and return a proper error
- Lines 126-133 & 140-145: We capture stdout/stderr in real-time for debugging

Plus, I think testing with binary process is better for e2e tests because:
- Running actual binaries tests the real deployment scenario - this is exactly how users will run the services, which will allow testing CLI parsing and proposer initialization that wouldn't be tested with proposer.run().

fault-proof/tests/common/contracts.rs
fakedev9999: @yuwen01 For the AccessManager, the problem is a combination of two things:
1. Ownable Contract: AccessManager inherits OpenZeppelin's Ownable, which sets the contract owner to msg.sender inside its constructor.
2. Anvil Fork Bug: Anvil's gas estimator struggles to correctly calculate the gas needed for this specific type of constructor logic when running on a forked network.

AnchorStateRegistry succeeds because it follows the proxy pattern. Its core logic isn't in a constructor but in an initialize function that is called after deployment. This basic proxy deployment is handled correctly with Anvil's gas estimator.

fault-proof/tests/common/monitor.rs
fakedev9999: Don't think we have a nice way since forge bind doesn't generate bindings for enums. See https://github.com/foundry-rs/foundry/issues/10441.

fault-proof/tests/common/monitor.rs
fakedev9999: Don't think we have a nice way since forge bind doesn't generate bindings for enums. See https://github.com/foundry-rs/foundry/issues/10441.

fault-proof/src/lib.rs
```diff
@@ -668,26 +679,13 @@ where
     }
 
     /// Attempts to resolve games, up to `max_games_to_check_for_resolution`.
-    #[tracing::instrument(
```
yuwen01, line 671: there were duplicated instruments, very pointless

contracts/foundry.toml
```diff
@@ -23,8 +23,8 @@ remappings = [
     "@lib-keccak/=lib/lib-keccak/contracts/lib",
 ]
 
-[lint]
-severity = ["high", "med", "low", "gas"] # Drop all "info" notes
+# [lint]
```
yuwen01, line 26: this causes an error with the forge script, where you can no longer access the config json files. I don't know why

contracts/script/fp/DeployOPSuccinctFDG.s.sol
```diff
@@ -29,19 +30,49 @@ import {Utils} from "../../test/helpers/Utils.sol";
...
+    struct DeployedContracts {
+        address factoryProxy;
+        address gameImplementation;
+        address sp1Verifier;
+        address anchorStateRegistry;
+        address accessManager;
+        address optimismPortal2;
+    }
+
+    function run()
```
yuwen01, line 44: returning this huge tuple looks troll, but it makes parsing the returns much cleaner

utils/client/src/client.rs
```diff
@@ -128,7 +128,7 @@ where
                     // Strip out all transactions that are not deposits.
                     attributes.transactions = attributes.transactions.map(|txs| {
```
yuwen01, line 129: clippy got mad

yuwen01 (approved):

contracts/script/fp/DeployOPSuccinctFDG.s.sol
ctian1: Are you not able to directly return the struct?

fault-proof/tests/common/env.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        // Put the test config into ../contracts/opsuccinctfdgconfig.json
+        let test_config: FaultDisputeGameConfig =
+            test_config(anvil.starting_l2_block_number, anvil.starting_root.clone());
+        let json = serde_json::to_string_pretty(&test_config)?;
+        let config_path = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+            .parent()
+            .unwrap()
+            .join("contracts")
+            .join("opsuccinctfdgconfig.json");
```
ctian1: nit: might be better to not hardcode this magic path here. Where else is it used?

ctian1 (approved):

PR #552, opened 2025-07-04
feat: validate agg

yuwen01:


validity/src/proposer.rs
```diff
@@ -623,6 +643,107 @@ where
...
+        debug!(
+            "Validating aggregation proof request: start_block={}, end_block={}",
+            agg_request.start_block, agg_request.end_block
+        );
+
+        // Fetch all completed range proofs within the aggregation range
+        let range_proofs = self
+            .driver_config
+            .driver_db_client
+            .get_consecutive_complete_range_proofs(
```
vcastellm, line 661: I don't see why are we retrieving the complete range proofs if we're validating `fetch_unrequested_agg_proof`? 🤔

validity/src/proposer.rs
```diff
@@ -623,6 +657,80 @@ where
...
+
+        // Log all constituent range proofs
+        for (i, proof) in range_proofs.iter().enumerate() {
+            debug!(
+                "Range proof {}: start_block={}, end_block={}",
+                i, proof.start_block, proof.end_block
+            );
+        }
+
+        // If no range proofs found, validation fails
```
vcastellm, line 679: According to https://github.com/succinctlabs/op-succinct/blob/main/validity/src/proposer.rs#L392 this can not happen

validity/src/proposer.rs
```diff
@@ -623,6 +654,104 @@ where
         Ok(None)
     }
 
+    /// Validates an aggregation proof request by checking that:
+    /// 1. There are no gaps between consecutive range proofs
+    /// 2. There are no duplicate/overlapping range proofs
```
fakedev9999, line 659: What if there exists duplicated range proofs? Will agg proof including that range never be requested for proof generation?

validity/src/proposer.rs
```diff
@@ -623,6 +654,104 @@ where
...
+                expected_end_block = ?agg_request.end_block,
+                actual_end_block = ?last_range_proof_request.end_block,
+                commitments = ?self.program_config.commitments,
+                "Range proofs end block does not match aggregation request"
+            );
+            return false;
+        }
+
+        // Check for gaps and duplicates / overlaps between consecutive proofs
+        for i in 1..range_proofs.len() {
```
fakedev9999, line 718: How about something like
```rust
let is_contiguous = range_proofs
        .windows(2)
        .all(|window| window[0].end_block == window[1].start_block);
```

validity/src/proposer.rs
yuwen01: then the overlap check will trigger

validity/src/proposer.rs
yuwen01: this doesn't differentiate between gaps and overlaps. irl, gaps will be much more common.

validity/src/proposer.rs
fakedev9999: @yuwen01 I know but won't the proposer be stuck and not be able to generate agg proof forever?

yuwen01: so basically, you are suggesting that this function should also be responsible for fixing the db if it's in a bad state? that hasn't happened in prod yet, and my intuition is that the db interactions aren't complicated enough where this is a failure case we need to prepare for, though I would be happy to work on this in a separate pr.

the exact error we encountered didn't corrupt the db at all. There are already a lot of checks to make sure there is only one writer to the DB. And when we insert range requests, we also double check to make sure that there are no duplicate requests. For these reasons, I feel like added resilience here may not be necessary, feel free to push back

fakedev9999: I think logging is enough for now :)

fakedev9999 (approved):
LGTM

PR #561, opened 2025-07-15
feat: update to support celestia indexer changes

fakedev9999:
There was an issue in the proposer for celestia da where batcher falls back to eth da for batch submission, the proposer cannot calculate safe l1 head for celestia. This was due to op-celestia indexer missing feature to handle eth da batches, only handling celestia da batches. Fix was made for op-celestia indexer in this [PR]( https://github.com/celestiaorg/optimism/pull/502), which introduces an API change.

This PR updates op-succinct's implementation according to the recent API change to calculate safe l1 head properly even the batcher posts batches to eth da on fallback.

### DIFF
--- a/fault-proof/bin/challenger.rs
+++ b/fault-proof/bin/challenger.rs
@@ -26,10 +26,10 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    setup_logger();
-
     let args = Args::parse();
     dotenv::from_filename(args.env_file).ok();
+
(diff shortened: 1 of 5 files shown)


fault-proof/bin/challenger.rs
```diff
@@ -26,11 +26,11 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    setup_logger();
-
```
yuwen01, line 30: oops

PR #572, opened 2025-07-25
feat: support eigenda

fakedev9999:
Bumps sp1, hana, hokulea for kona-client/v1.0.2 with MSRV 1.86.


programs/range/eigenda/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+use op_succinct_range_utils::setup_tracing;
+use rkyv::rancor::Error;
+
+fn main() {
+    #[cfg(feature = "tracing-subscriber")]
+    setup_tracing();
+
+    kona_proof::block_on(async move {
+        let witness_rkyv_bytes: Vec<u8> = sp1_zkvm::io::read_vec();
+        let witness_data = rkyv::from_bytes::<EigenDAWitnessData, Error>(&witness_rkyv_bytes)
```
ctian1, line 29: EigenDAWitnessData contains the full canoe sp1 compressed proof bytes which is >1 MB. This might be adding significant unnecessary cycles?

resources/g1.point
ctian1, line 1: File looks unused?

utils/eigenda/host/src/witness_generator.rs
```diff
@@ -0,0 +1,157 @@
...
+            &eigenda_witness_data,
+            oracle.clone(),
+            canoe_provider,
+        )
+        .await?;
+
+        // Store the canoe proof in the witness data
+        let canoe_proof_bytes =
+            serde_cbor::to_vec(&canoe_proofs).expect("Failed to serialize canoe proof");
+        eigenda_witness_data.canoe_proof_bytes = Some(canoe_proof_bytes);
```
ctian1, line 168: Seems this shouldn't go into the witness that is read into the VM

https://github.com/Layr-Labs/hokulea/blob/master/crates/proof/src/canoe_verifier/sp1_cc.rs#L53

resources/g1.point
fakedev9999: Yeah that's for running the eigenda proxy. Will delete.

programs/range/eigenda/src/main.rs
fakedev9999: WDYM "unnecessary"? are there clever ways to avoid this?

utils/eigenda/host/src/witness_generator.rs
fakedev9999: Made a fix [here](https://github.com/succinctlabs/op-succinct/pull/572/commits/ec638a47c8d6fa44f47b1202c2a79e222628c0db)

Cargo.toml
```diff
@@ -100,6 +102,14 @@ hana-celestia = { git = "https://github.com/celestiaorg/hana", tag = "v0.1.0-bet
 hana-host = { git = "https://github.com/celestiaorg/hana", tag = "v0.1.0-beta.17" }
 hana-oracle = { git = "https://github.com/celestiaorg/hana", tag = "v0.1.0-beta.17" }
 
+# hokulea
+hokulea-eigenda = { git = "https://github.com/Layr-Labs/hokulea", rev = "ab9d4f7" }
```
bxue-l2: use a more updated version

programs/range/eigenda/src/main.rs
bxue-l2: I don't think you want to put the sp1 compressed proof into the zkVM along with other data via stdin.
The proof should have been transmitted "write_proof".

https://github.com/Layr-Labs/hokulea/blob/0a6200cd0f22caa28aca040e47860dd42893ae26/crates/proof/src/eigenda_blob_witness.rs#L56

utils/client/src/witness/mod.rs
```diff
@@ -58,6 +58,26 @@ impl WitnessData for DefaultWitnessData {
     }
 }
 
+#[derive(Clone, Debug, Default, rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
+pub struct EigenDAWitnessData {
+    pub preimage_store: PreimageStore,
+    pub blob_data: BlobData,
+    // EigenDABlobWitnessData.
+    // See https://github.com/Layr-Labs/hokulea/blob/056ee9888964f1a63191384ce882a2d4f805e76e/crates/proof/src/eigenda_blob_witness.rs.
```
bxue-l2: a newer version?

utils/eigenda/host/src/witness_generator.rs
```diff
@@ -0,0 +1,165 @@
...
+            WitnessExecutorTrait::run(&executor, boot_info.clone(), pipeline, cursor, l2_provider)
+                .await
+                .unwrap();
+        }
+
+        // Extract the EigenDA witness data
+        let mut eigenda_witness_data = std::mem::take(&mut *eigenda_blobs_witness.lock().unwrap());
+
+        // Generate canoe proofs using the reduced proof provider for proof aggregation
+        use canoe_sp1_cc_host::CanoeSp1CCReducedProofProvider;
```
bxue-l2, line 149: check if there is no DAcerts for the range to prove, skip this step entirely?

programs/range/eigenda/src/main.rs
ctian1: Fixed [here](https://github.com/succinctlabs/op-succinct/pull/572/files#diff-be3d2b84170ba0f06740dc3f6d850c35aa881c5f1cf5937198818bd128bb8f25R56) already

resources/g1.point
ctian1: Still needs to be removed

PR #575, opened 2025-08-01
feat: audit fix

yuwen01:


Cargo.toml
```diff
@@ -127,7 +127,9 @@ alloy-network = { version = "1.0.9" }
 
 # Alloy
 alloy-rlp = { version = "0.3.12", default-features = false }
-alloy-trie = { version = "0.8.1", default-features = false, features = ["ethereum"] }
+alloy-trie = { version = "0.8.1", default-features = false, features = [
```
fakedev9999, line 130: Oh are you using some formatter

contracts/src/validity/OPSuccinctL2OutputOracle.sol
```diff
@@ -387,11 +426,11 @@ contract OPSuccinctL2OutputOracle is Initializable, ISemver {
...
     {
-        // The proposer must be explicitly approved, or the zero address must be approved (permissionless proposing),
-        // or the fallback timeout has been exceeded allowing anyone to propose.
+        // The proposer must be explicitly approved, or the zero address must be approved (permissionless proposing).
+        // In optimistic mode, there is NO permissionless fallback timeout. That is, no matter how long it has been since the last proposal,
+        // non-permissioned proposers cannot propose in optimistic mode.
         require(
-            approvedProposers[msg.sender] || approvedProposers[address(0)]
-                || (block.timestamp - lastProposalTimestamp() > fallbackTimeout),
+            approvedProposers[msg.sender] || approvedProposers[address(0)],
```
fakedev9999, line 433: What's your thoughts on removing proposeL2Output for whenOptimistic? Or at least I think we should check with l2beat whether not having permissionless fallback timeout in optimistic mode might be a blocker for achieving stage 1.

contracts/src/validity/OPSuccinctL2OutputOracle.sol
```diff
@@ -433,6 +472,33 @@ contract OPSuccinctL2OutputOracle is Initializable, ISemver {
...
+        bytes32 _outputRoot,
+        uint256 _l2BlockNumber,
+        uint256 _l1BlockNumber,
+        bytes memory _proof,
+        address _proverAddress
+    ) external payable whenNotOptimistic returns (IDisputeGame _game) {
+        require(disputeGameFactory != address(0), "L2OutputOracle: dispute game factory is not set");
+
+        _enteredDGFCreate = true;
+        _game = IDisputeGameFactory(disputeGameFactory).create{value: msg.value}(
```
fakedev9999, line 496: I understand how it works but the cycle where GameImpl wraps over L2OO but the game is created by calling L2OO which calls the factory.create seems like an anti pattern but I cannot think of a better way to fix the issue.

validity/src/proposer.rs
```diff
@@ -925,21 +924,16 @@ where
...
-                completed_agg_proof.proof.as_ref().unwrap().clone().into(),
-            ));
-
             let transaction_request = self
                 .contract_config
-                .dgf_contract
-                .create(
-                    OP_SUCCINCT_VALIDITY_DISPUTE_GAME_TYPE,
+                .l2oo_contract
+                .dgfProposeL2Output(
```
fakedev9999, line 930: How about adding a note here not to modify the binary to use the standard interface to do `dgf_contract.create` in permissionless mode due to front running attacks? Or at least document it somewhere in the book?

PR #577, opened 2025-08-02
chore: unify binary name for celestia docker image

fakedev9999:
Align the Celestia Docker image binary name with the standard validity proposer to ensure consistent deployment across all variants. This reduces complexity of infra code switching between op-succinct and op-succinct-celestia images.

Docker image tag already differentiates the celestia variant, so having different binary name seems to just increase complexity.


vcastellm (approved):
LGTM

yuwen01 (approved):

PR #579, opened 2025-08-05
feat: add cycles + prover gas metrics

fakedev9999:
Adding more game proving stats for op-succinct-lite proposer to track performance.
Also improved logging for better visibility and debugging.

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -133,7 +133,7 @@ where
     }
 
     #[tracing::instrument(name = "[[Proving]]", skip(self), fields(game_address = ?game_address))]
-    pub async fn prove_game(&self, game_address: Address) -> Result<TxHash> {
+    pub async fn prove_game(&self, game_address: Address) -> Result<(TxHash, u64, u64)> {
         tracing::info!("Attempting to prove game {:?}", game_address);
 
         let fetcher = match OPSuccinctDataFetcher::new_with_rollup_config().await {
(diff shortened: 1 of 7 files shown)


fault-proof/src/proposer.rs
```diff
@@ -133,7 +133,7 @@ where
     }
 
     #[tracing::instrument(name = "[[Proving]]", skip(self), fields(game_address = ?game_address))]
-    pub async fn prove_game(&self, game_address: Address) -> Result<TxHash> {
+    pub async fn prove_game(&self, game_address: Address) -> Result<(TxHash, u64, u64)> {
```
yuwen01, line 143: add a doc comment here about what the u64s mean? in my experience outputting tuples gets confusing fast

fault-proof/src/proposer.rs
```diff
@@ -171,28 +171,48 @@ where
         };
 
         tracing::info!("Generating Range Proof");
-        let range_proof = if self.config.mock_mode {
+        let (range_proof, total_instruction_cycles, total_sp1_gas) = if self.config.mock_mode {
             tracing::info!("Using mock mode for range proof generation");
-            let (public_values, _) =
+            let (public_values, report) =
                 self.prover.network_prover.execute(get_range_elf_embedded(), &sp1_stdin).run()?;
```
yuwen01: can we just add a note here that gas is counted by default? or do 
```suggestion
              self.prover.network_prover.execute(get_range_elf_embedded(), &sp1_stdin).calculate_gas(true).run()?;
```

yuwen01 (approved):

PR #582, opened 2025-08-07
fix: aggregated ranges must be contiguous

### LINKED ISSUE #581: Validity proposer can request invalid aggregation proofs
This was a gnarly one to debug.

The validity proposer is susceptible to a race condition. In the proposer, `get_highest_proven_contiguous_block` does not actually enforce that the block is contiguous.

I ran into this on the Sigil testnet when running multiple provers of varied speeds creates range proofs with very different timings. Example:

`| 0 - 3200 -> L2OO | 3200 - 3300 -> pending | 3300 - 3400 -> done | 3400 - 3500 -> done |`

My validity proposer created an aggregation proof of 3400-3600 before trying to submit it. Once that proof was in the database, it gummed up the works forever and was unable to verify on the smart contract because it necessarily had an `l2PreRoot` that did not match block 3200.

TimTinkers:
I have modified `get_highest_proven_contiguous_block` to require knowing the `latest_proposed_block_number` to confirm whether a range is actually contiguous or not.

This fixes #581.

### DIFF
--- a/validity/src/proposer.rs
+++ b/validity/src/proposer.rs
@@ -487,7 +487,10 @@ where
 
         // Get the highest block number of the completed range proofs.
         let highest_proven_contiguous_block_number = match self
-            .get_highest_proven_contiguous_block(completed_range_proofs)?
+            .get_highest_proven_contiguous_block(
+                latest_proposed_block_number,
+                completed_range_proofs
+            )?
         {
             Some(block) => block,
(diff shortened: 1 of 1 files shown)


fakedev9999:
This issue was fixed in [this PR](https://github.com/succinctlabs/op-succinct/pull/552), which is included in the v3.0.0 release.

PR #585, opened 2025-08-09
fix: use end exclusive range

fakedev9999:
Fixes to treat Blobstream endBlock as exclusive in commitment check.

Blobstream DataCommitmentStored ranges are [start, end), but the original code treated `endBlock` as inclusive.
When target_celestia_height == endBlock, we incorrectly matched the *previous* commitment and returned its L1 block,
potentially causing witness generation to later fail with "No matching event found for the given Celestia height" error.

Changed the upper-bound check to `< endBlock` and update logs to print the range as `[start, end)` for clarity.

This should fix to select the correct (later) commitment and L1 head when height hits a boundary.

### DIFF
--- a/scripts/utils/bin/fetch_fault_dispute_game_config.rs
+++ b/scripts/utils/bin/fetch_fault_dispute_game_config.rs
@@ -1,5 +1,6 @@
 use std::{env, sync::Arc};
 
+use alloy_eips::BlockId;
 use anyhow::Result;
 use fault_proof::config::FaultDisputeGameConfig;
 use op_succinct_host_utils::{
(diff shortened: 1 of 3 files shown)


scripts/utils/bin/fetch_fault_dispute_game_config.rs
```diff
@@ -70,6 +72,7 @@ use std::env;
 /// needed for the Solidity deployment scripts.
 async fn update_fdg_config() -> Result<()> {
```
yuwen01, line 74: need to change this in `fetch_l2oo_config.rs` too?

yuwen01 (approved):

PR #590, opened 2025-08-19
docs: make update parameters flow more clear

yuwen01:


book/validity/contracts/update-parameters.md
```diff
@@ -2,43 +2,18 @@
 
 If you just need to update the `aggregationVkey`, `rangeVkeyCommitment` or `rollupConfigHash` parameters and not upgrade the contract itself, you can use the `just add-config` and `just remove-config` command.
```
0xernesto: I feel like the beginning of this page is missing context. We should make it clear why one would end up in a situation where these would need to be updated.

From my understanding, you end up in this situation if you make changes to the proposer code or something that requires updating the ELFs. Right?

book/validity/contracts/update-parameters.md
```diff
@@ -2,43 +2,18 @@
...
-just remove-config <config_name>
-```
-
-### Permissions
-
-#### Using an EOA `ADMIN_PK` key
+## Rolling upgrade guide
 
-If you are the owner of the `OPSuccinctL2OutputOracle` contract, you can set `ADMIN_PK` in your `.env` to directly add and remove configurations. If unset, this will default to the value of `PRIVATE_KEY`.
+1. Perform some update that changes your `aggregationVkey`, `rangeVkeyCommitment` or `rollupConfigHash` locally. For more information on producing the elfs and vkeys, see [this page](../../advanced/verify-binaries.md).
```
0xernesto: I would change this to this:

```md
1. Generate new elfs and vkeys, see [this page](../../advanced/verify-binaries.md).
```

PR #591, opened 2025-08-19
refactor: start range proof from the database instead

fakedev9999:
Fixes the range proof starting point use the latest successful range from the database instead of the contract's latest block.

This change eliminates duplicate work by not creating requests for already-proven ranges, but still uses the contract block if no database proofs exist.

### DIFF
--- a/validity/src/proposer.rs
+++ b/validity/src/proposer.rs
@@ -158,22 +158,56 @@ where
         Ok(proposer)
     }
 
-    /// Use the in-memory index of the highest block number to add new ranges to the database.
+    /// Use the highest proven contiguous block from the database to add new ranges to the database.
     #[tracing::instrument(name = "proposer.add_new_ranges", skip(self))]
     pub async fn add_new_ranges(&self) -> Result<()> {
         // Get the latest proposed block number on the contract.
-        let latest_proposed_block_number = get_latest_proposed_block_number(
+        let contract_latest_proposed_block_number = get_latest_proposed_block_number(
             self.contract_config.l2oo_address,
             self.driver_config.fetcher.as_ref(),
         )
         .await?;
+
+        // Get the highest proven contiguous block from the database to use as our starting point.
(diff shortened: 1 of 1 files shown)


validity/src/proposer.rs
```diff
@@ -158,22 +158,56 @@ where
...
 
+        // Get the highest proven contiguous block from the database to use as our starting point.
+        // This prevents creating duplicate range requests when the database has proofs beyond
+        // what's been proposed to the contract.
+        let completed_range_proofs = self
+            .driver_config
+            .driver_db_client
+            .fetch_completed_ranges(
+                &self.program_config.commitments,
+                0, // Start from block 0 to get all completed ranges
```
0xernesto, line 179: should this be `contract_latest_proposed_block_number`?

0xernesto: Also, do we need to sort `completed_range_proofs`?

validity/src/proposer.rs
```diff
@@ -158,22 +158,56 @@ where
...
+                self.requester_config.l1_chain_id,
+                self.requester_config.l2_chain_id,
+            )
+            .await?;
+
+        // Use the highest proven contiguous block as our starting point, falling back to contract's
+        // latest block
+        let starting_block_number =
```
0xernesto, line 187: Maybe do this so we never start below the contract latest block.

```rust
let starting_block_number = match self.get_highest_proven_contiguous_block(completed_range_proofs)? {
    Some(h) => std::cmp::max(h as u64, contract_latest_proposed_block_number),
    None => contract_latest_proposed_block_number,
};
```

0xernesto (changes requested):

validity/src/proposer.rs
fakedev9999: `get_highest_proven_contiguous_block` expects the input range to be sorted

PR #595, opened 2025-08-21
feat: Allow to specify registry & factory to avoid re-deploying.

Mc01:
Upstreaming changes present in deployment of OpSuccinct to Celo networks: https://github.com/celo-org/op-succinct/pull/11

Introduced:
- `.env.example` as template for `.env`
- optional creation of registry, factory & superchain config
- extracted fetching of config into separate recipe in justfile
- introduced deploying of contracts with custom config without fetching in justfile


Mc01:
Since downstream diverged a lot - closing 😌 
Will reopen one day with more recent changes 😁

PR #597, opened 2025-08-22
feat: check ancestor chain on game creation

fakedev9999:
Adds a game chain validation logic to prevent game creation on invalid chain of games.
This includes (1) a game that's still in process but has invalid output root or (2) a game that has valid output root but has been resolved in favor of challenger.


ctian1 (commented):
- With these MAX_GAMES params, if someone were to make 101 malicious games at once, would the first one be ignored by default?
  - With some of these params, it’s not clear why they exist or when they should be changed.
  - Why is there max_depth_to_check? Could someone attack by creating fake chain and then 100 valid proposals after it or something?
- So now the proposer needs to make up to 300 RPC calls on every loop to ensure the latest game is valid to build on top of? Could have some performance / RPC cost concerns there. Can we optimize the happy path by cacheing the latest game we proposed locally? Or, having an LRUCache of recently proposed games, and finding a valid game that points whose parent is in the cache.

PR #598, opened 2025-08-26
docs: L2 node latency + cleanup

0xernesto:
- Add warning about latency issues when L2 nodes and proposer are not in the same region/network
- Miscellaneous cleanup and improvements


fakedev9999 (approved):
LGTM

PR #606, opened 2025-09-04
fix(contracts): upgrade, missing proxy in calldata for proxy admin managed instance (not naked proxies)

fakedev9999:


Hamzaraza351:
Commit 2.02

PR #614, opened 2025-09-17
feat: spawn defense tasks concurrently

leruaa:
Get all defensible game addresses and spawn a task for all of them, instead of doing it for the oldest only.


fault-proof/src/proposer.rs
```diff
@@ -835,25 +835,25 @@ where
...
+            .get_defensible_game_addresses(
                 self.config.max_games_to_check_for_defense,
                 self.l1_provider.clone(),
                 self.l2_provider.clone(),
             )
-            .await?
-        {
+            .await?;
+
+        let mut tasks_spanned = false;
```
```suggestion
        let mut tasks_spawned = false;
```

fakedev9999 (approved):
LGTM

PR #617, opened 2025-09-18
feat: add save artifacts flag for multi script

fakedev9999:
Adds `--save-artifacts` flag for multi script so that the program and stdin artifacts can be saved locally.


scripts/utils/src/lib.rs
```diff
@@ -30,6 +30,9 @@ pub struct HostExecutorArgs {
     /// Whether to generate proofs.
     #[arg(long)]
     pub prove: bool,
+    /// Whether to save the program and stdin artifacts locally.
+    #[arg(long)]
```
kevjue, line 34: Would it make sense to have a default value for this command line option?

PR #622, opened 2025-09-24
feat: get execution stats from network when not in mock mode

leruaa:
Get the execution stats from the network using `get_proof_request()`


fakedev9999 (approved):
lgtm

PR #624, opened 2025-09-26
fix: bump hokulea version and fix only ethda problem

bxue-l2:
This PR does three things

1. it bumps the hokulea version, and incorporate the rename PR https://github.com/Layr-Labs/hokulea/pull/160
2. it resolves the issue when there is only EthDA in the derivation pipeline. 
  - Previously the `EigenDAWitnessData::eigenda_data` is assigned to None, and the system would aborts at the client side expecting one. 
  - Now, `EigenDAWitnessData::eigenda_data` is always assigned to something, and sometimes it can be an empty struct
3. it incorporates the new hokulea change, where `hokulea_witgen::from_boot_info_to_canoe_proof` now returns an option, and when the returned value is None, it indicates there is no canoe proof to be expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2595,7 +2595,7 @@ dependencies = [
 [[package]]
 name = "canoe-bindings"
 version = "0.1.0"
-source = "git+https://github.com/Layr-Labs/hokulea?rev=2dae748#2dae748c69d7ace01ab1c1ff471e87449371893d"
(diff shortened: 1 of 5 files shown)


fakedev9999 (approved):
lgtm

fakedev9999:
Merging from #626 and closing due to github secrets ci issue.

PR #635, opened 2025-10-07
refactor: refactor run_range_program

fakedev9999:
Refactors the shared `run_range_program` helper to accept pre-built oracle and blob store instances, preventing duplicate work for eigenda range program.

### DIFF
--- a/book/advanced/cost-estimator.md
+++ b/book/advanced/cost-estimator.md
@@ -20,7 +20,6 @@ Before running the cost estimator, you need to set up a .env file in the project
 
 ```bash
 L1_RPC=<YOUR_L1_RPC_ENDPOINT>
-L1_BEACON_RPC=<YOUR_L1_BEACON_RPC_ENDPOINT>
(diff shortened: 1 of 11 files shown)


programs/range/ethereum/src/main.rs
```diff
@@ -25,6 +25,11 @@ fn main() {
         let witness_data = rkyv::from_bytes::<DefaultWitnessData, Error>(&witness_rkyv_bytes)
             .expect("Failed to deserialize witness data.");
 
-        run_range_program(ETHDAWitnessExecutor::new(), witness_data).await;
+        let (oracle, beacon) = witness_data
+            .get_oracle_and_blob_provider()
```
kevjue, line 29: Looks like there is a clippy failure here.

fakedev9999: Fixed.

kevjue (approved):
LGTM!

PR #639, opened 2025-10-10
fix: add timeouts to network prover calls to prevent deadlock

fakedev9999:
## Problem
The proposer could freeze indefinitely when network prover calls hang without timeouts.

## Solution
Wrap all NetworkProver API calls (get_proof_status, get_proof_request, cancel_request) with 60-second timeouts. When 
timeouts occur, error context is logged and the proposer loop continues instead of hanging silently.

### DIFF
--- a/validity/src/proposer.rs
+++ b/validity/src/proposer.rs
@@ -35,7 +35,7 @@ use crate::{
 };
 
 /// Timeout for network prover calls to prevent indefinite hangs.
-const NETWORK_CALL_TIMEOUT_SECS: u64 = 60;
+const NETWORK_CALL_TIMEOUT_SECS: u64 = 15;
 
 /// Configuration for the driver.
 pub struct DriverConfig {


validity/src/proposer.rs
```diff
@@ -1513,4 +1536,62 @@ where
...
+                    "Timeout after {}s {} for request {} (start_block={}, end_block={})",
+                    NETWORK_CALL_TIMEOUT_SECS,
+                    operation_name,
+                    request.id,
+                    request.start_block,
+                    request.end_block
+                ))
+            }
+        }
+    }
```
borngraced, line 1596: thought of retry too?

validity/src/proposer.rs
```diff
@@ -34,6 +34,9 @@ use crate::{
     ProgramConfig, RequestExecutionStatistics, RequesterConfig, ValidityGauge,
 };
 
+/// Timeout for network prover calls to prevent indefinite hangs.
+const NETWORK_CALL_TIMEOUT_SECS: u64 = 60;
```
kevjue: Did you ask the network team if this is a reasonable value?

validity/src/proposer.rs
kevjue: Related to this question, what happens when a timeout occurs without any retries?

fakedev9999: @kevjue As I mentioned from the call yesterday, the row will be deleted from the db, re-inserted and start over from witness generation.

validity/src/proposer.rs
fakedev9999: I'll ask

fakedev9999: Confirmed.

fakedev9999: Oh Matt told me network team usually does 15s. Will fix to 15s.

validity/src/proposer.rs
kevjue: Sounds good.  So its effectively going to retry.  Thanks for the clarification.

kevjue (approved):
optimistically approving.  Please don't forget to make that timeout change.

PR #641, opened 2025-10-14
fix: validate cached checkpoints against contract before reuse

fakedev9999:
Validity proposer now validate cached checkpoints against contract before reuse to recover from L1 reorgs.

## Changes 
- expose historicBlockHashes(uint256) on the L2OO binding so the proposer can query the live mapping
- compare cached aggregation checkpoints with the contract’s value and fall back to re-checkpointing when they diverge


kevjue (approved):
LGTM

Issue #642, opened 2025-10-14
Make `L1_BEACON_RPC` optional


hai-rise:
We are trying to integrate OP Succinct Lite into our custom OP Stack, and don't think we need `L1_BEACON_RPC` anywhere in our flow. The Challenger would panic without one, though. Making it optional or even removing it would help simplify integrations (especially for devnets) 🤔. 

There is already this in-code comment as well:

https://github.com/succinctlabs/op-succinct/blob/52d5d842aba25e41ca5df138f41a9fedcdfe9571/utils/host/src/fetcher.rs#L52-L53

PR #645, opened 2025-10-15
feat: optional `L1_BEACON_RPC`

Farhad-Shabani:
Resolves: #642

### DIFF
--- a/utils/host/src/fetcher.rs
+++ b/utils/host/src/fetcher.rs
@@ -200,8 +200,8 @@ impl OPSuccinctDataFetcher {
                         // tx.inner.effective_gas_price * tx.inner.gas_used +
                         // tx.l1_block_info.l1_fee is the total fee for the transaction.
                         // tx.inner.effective_gas_price * tx.inner.gas_used is the tx fee on L2.
-                        tx.inner.effective_gas_price * tx.inner.gas_used as u128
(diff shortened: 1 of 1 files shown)


utils/host/src/fetcher.rs
```diff
@@ -288,12 +295,12 @@ impl OPSuccinctDataFetcher {
     }
 
     /// Get the RPC URL for the given RPC mode.
-    pub fn get_rpc_url(&self, rpc_mode: RPCMode) -> &Url {
+    pub fn get_rpc_url(&self, rpc_mode: RPCMode) -> Option<&Url> {
```
leruaa: How about return a `Result` instead an `Option`? that would allow to return a more specific error message for L1

Farhad-Shabani: I generally (library-wise) would prefer to keep methods agnostic about what constitutes an error. Simply return the most natural type for the data, in this case, an `Option`. Allowing the caller to decide whether the absence of a value is acceptable or should be treated as an error.

That said, since this method is only used internally here, and other similar methods already follow a `Result`-based pattern, it makes sense to align. [41e0f8b](https://github.com/succinctlabs/op-succinct/pull/645/commits/41e0f8badc72cd967790057483facafc632242fd)

leruaa (approved):
lgtm

PR #653, opened 2025-10-21
fix(fault-proof): clear stale canonical head value

fakedev9999:
## Problem
After a game was successfully challenged and resolved as CHALLENGER_WINS, the proposer incorrectly kept trying to create its next game using this invalid game as the parent. This caused a persistent InvalidParentGame() error loop, which prevented the proposer from recovering automatically and required a manual restart to resume creating games from a valid parent.

The root cause was due to state inconsistency in compute_canonical_head() where stale canonical_head_index persists after all valid games are removed from cache.

## Solution 

Clear stale canonical head values in compute_canonical_head

```diff
if let Some(canonical_head) = canonical_head {
     state.canonical_head_index = Some(canonical_head.index);
     state.canonical_head_l2_block = Some(canonical_head.l2_block);
- }
+ } else {
+    // Clear stale canonical head when no valid games exist
+    state.canonical_head_index = None;
+ }
```

## TODO
- [x] Add e2e test to cover this case

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -896,28 +896,12 @@ where
     }
 
     /// Handles the creation of a new game if conditions are met.
-    /// Returns the address of the created game, if one was created.
     #[tracing::instrument(name = "[[Proposing]]", skip(self))]
-    pub async fn handle_game_creation(&self) -> Result<Option<Address>> {
-        let (latest_proposed_block_number, parent_game_index) = {
-            let state = self.state.lock().await;
-
-            let Some(latest_proposed_block_number) = state.canonical_head_l2_block else {
-                tracing::info!("No canonical head; skipping game creation");
-                return Ok(None);
-            };
-
-            let parent_game_index =
-                state.canonical_head_index.map(|index| index.to::<u32>()).unwrap_or(u32::MAX);
-
-            (latest_proposed_block_number, parent_game_index)
-        };
-
-        // Choose the next L2 block number for proposal.
-        // If there already exists a game at the next L2 block number for proposal, increment the L2
-        // block number by 1
-        let mut next_l2_block_number_for_proposal =
-            latest_proposed_block_number + U256::from(self.config.proposal_interval_in_blocks);
+    pub async fn handle_game_creation(
+        &self,
+        mut next_l2_block_number_for_proposal: U256,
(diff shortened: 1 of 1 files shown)


fault-proof/src/proposer.rs
```diff
@@ -472,9 +474,32 @@ where
...
+                    previous_canonical_index = ?previous_canonical_index,
+                    new_canonical_index = %canonical_head.index,
+                    l2_block = %canonical_head.l2_block,
+                    total_games = state.games.len(),
+                    "Canonical head updated"
+                );
+            }
+        } else {
+            // Clear stale canonical head index when no valid games exist.
+            state.canonical_head_index = None;
```
Farhad-Shabani, line 494: So just to confirm: is it possible for the store to have no game at all, but still have `canonical_head_index` set to a value?

fault-proof/src/proposer.rs
```diff
@@ -901,8 +913,37 @@ where
             (latest_proposed_block_number, parent_game_index)
         };
 
-        let next_l2_block_number_for_proposal =
+        // Choose the next L2 block number for proposal.
+        // If there already exists a game at the next L2 block number for proposal, increment the L2
+        // block number by 1
+        let mut next_l2_block_number_for_proposal =
             latest_proposed_block_number + U256::from(self.config.proposal_interval_in_blocks);
```
Farhad-Shabani: I see one level up before calling `handle_game_creation`, we are obtaining the `next_l2_block_number_for_proposal`. Wondering couldn't we pass that through and rely on? so not having to recompute?

Also not sure if I get the cases where a game can exist for the next L2 block number for proposal.

fault-proof/src/proposer.rs
fakedev9999: Discussed IRL.

TL;DR is yes, the previous code was bugged and was possible to land in a case where the cache is empty yet still have canonical_head_index set to a value.

fault-proof/src/proposer.rs
fakedev9999: I want to keep as-is since the "redundant" computation gives safety ensuring not to error in case there's inconsistency in the cache between the time when should_create_game is called and game creation task is spawned in the background.

Farhad-Shabani (approved):
✨

Issue #655, opened 2025-10-22
Add test coverage reporting to CI


Farhad-Shabani:
Integrate `cargo-llvm-cov` into our CI pipeline to collect and report test coverage.

Tasks:
- [ ] Install and configure `cargo-llvm-cov` in CI
- [ ] Enforce minimum coverage threshold (?)
- [ ] Generate report and upload to Codecov
- [ ] Add coverage badge to README

PR #657, opened 2025-10-23
ci: add test coverage reporting to CI

Farhad-Shabani:
Resolves: #655

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,0 +1,114 @@
+name: Coverage
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/coverage.yaml
+      - "Cargo.toml"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
     - run: rustup toolchain install stable --profile minimal
     - uses: Swatinem/rust-cache@v2  
     - name: Install SP1 toolchain
-        run: |
```
Farhad-Shabani: @fakedev9999, looks like the "Cargo Tests" were skipped because of this error. Now I see the [same error](https://github.com/succinctlabs/op-succinct/actions/runs/18739873750/job/53453789338?pr=657#step:8:2244) popping up in CI that I was talking about earlier when running things locally.

Issue #659, opened 2025-10-23
`validity::utils` docstring tests fail


Farhad-Shabani:
By running `cd validity/utils && cargo test --release`:

```bash
test validity/src/utils.rs - utils::find_gaps (line 12) ... FAILED
test validity/src/utils.rs - utils::get_ranges_to_prove_by_blocks (line 136) ... FAILED
```

Issue #658, opened 2025-10-23
Fault-proof tests race to spawn Anvil and panics leave Anvil running in background


Farhad-Shabani:
When running concurrent tests, each one tries to spawn a new Anvil instance at the same time, which causes port conflicts.
On top, if any test panics, the Anvil subprocess it spawned keeps running in the background.

There's also an issue with the CI (ci.yaml), it's skipping the tests entirely.

PR #660, opened 2025-10-23
fix: re-enable CI tests + resolve Anvil conflict + forge fmt

Farhad-Shabani:
Resolves: #658
Resolves: #659

Along with `forge fmt` with nightly (required by CI)

### DIFF
--- a/justfile
+++ b/justfile
@@ -336,8 +336,6 @@ tests:
     -- \
     --skip test_cycle_count_diff \
     --skip test_post_to_github \
-    --skip test_honest_proposer \
-    --skip test_honest_challenger
 
 # Run end-to-end tests.
 e2e-tests:


.github/workflows/ci.yaml
```diff
@@ -26,13 +26,14 @@ jobs:
...
-        run: |
+      run: |
           curl -L https://sp1.succinct.xyz | bash
           ~/.sp1/bin/sp1up 
           ~/.sp1/bin/cargo-prove prove --version
           source ~/.bashrc
+
     - name: Test
-      run: cargo test --release -- --skip test_cycle_count_diff --skip test_post_to_github --skip test_honest_proposer --skip test_honest_challenger
+      run: cargo test --release -- --skip test_cycle_count_diff --skip test_post_to_github --skip test_honest_proposer --skip test_honest_challenger --test-threads=1
```
Farhad-Shabani: I just wanted to get CI working for now, so I added the `--test-threads=1` flag. But we can probably revisit our test codebase later, then we can drop this.

fakedev9999: Can you add the rest of the e2e tests to skip?

Farhad-Shabani: I ended up putting the e2e tests behind a feature flag. Also, running it using `just`, so we can run locally while developing with the same command. Hopefully, it works now!

PR #666, opened 2025-10-29
feat: load .env from the specified path before falling back to root

vcastellm:
## Description

This is a follow-up on my last [PR](https://github.com/succinctlabs/op-succinct/pull/652), it will try to load the env file from the specified path or fallback to search in the root project.

This also allows the fetch-l2oo-config to work standalone.

> [!note]
> This change is backwards compatible

### DIFF
--- a/scripts/utils/bin/fetch_l2oo_config.rs
+++ b/scripts/utils/bin/fetch_l2oo_config.rs
@@ -152,13 +152,13 @@ struct Args {
 async fn main() -> Result<()> {
     let args = Args::parse();
 
-    // Load the env file in case it's present
-    if !dotenv::from_path(args.env_file.clone()).is_ok() {
-        // This fetches the .env file from the project root. If the command is invoked in the contracts/
-        // directory, the .env file in the root of the repo is used.
-        if let Some(root) = find_project_root() {
-            dotenv::from_path(root.join(args.env_file)).ok();
-        } else {
+    // This fetches the .env file from the project root. If the command is invoked in the contracts/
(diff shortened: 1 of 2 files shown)


scripts/utils/bin/fetch_l2oo_config.rs
```diff
@@ -152,12 +152,15 @@ struct Args {
...
     let args = Args::parse();
 
-    // This fetches the .env file from the project root. If the command is invoked in the contracts/
-    // directory, the .env file in the root of the repo is used.
-    if let Some(root) = find_project_root() {
-        dotenv::from_path(root.join(args.env_file)).ok();
-    } else {
-        eprintln!("Warning: Could not find project root. {} file not loaded.", args.env_file);
+    // Load the env file in case it's present
+    if !dotenv::from_path(args.env_file.clone()).is_ok() {
```
Farhad-Shabani:
```suggestion
    if dotenv::from_path(args.env_file.clone()).is_err() {
```

Farhad-Shabani:
@vcastellm could you also make the first load (default one) from the `root`? if not found then try `dotenv::from_path(args.env_file.clone())`?

vcastellm:
Thanks for the feedback @Farhad-Shabani, I think it's better now

Farhad-Shabani:
Of course @vcastellm 
Closing this and reopening as #667 so all CI jobs can run.
Also it might stay a bit. We are seeing an issue on `head`, will be fixing that first, then merge this one.

PR #669, opened 2025-10-30
feat: add preflight check script for fp

fakedev9999:
Adds a preflight check script for OPSuccinctFaultDistputeGame to catches configuration issues before production by testing the complete proof generation and submission pipeline end-to-end. 

Note that the game creation and proof submission is simulated on a forked L1 network to not touch the actual storage on mainnet.

### DIFF
--- a/book/introduction.md
+++ b/book/introduction.md
@@ -13,7 +13,7 @@ It combines a few key technologies:
 - [SP1](https://docs.succinct.xyz/docs/sp1/introduction), Succinct's state-of-the-art Rust zkVM
 - [Succinct Prover Network](https://docs.succinct.xyz/docs/sp1/prover-network/intro), Succinct's low-latency, cost-effective proving API
 
-OP Succinct is the only production-ready proving solution for the OP Stack and trusted by teams like [Mantle](https://www.mantle.xyz/blog/announcements/op-succinct-mantle-network-testnet) and [Phala](https://phala.network/posts/phala-network-20-first-opsuccinct-layer-2-on-ethereum).
+OP Succinct is the only production-ready proving solution for the OP Stack and trusted by teams like [Mantle](https://www.mantle.xyz/blog/announcements/op-succinct-mantle-network-testnet) and [Phala](https://phala.com/).
 
 
 ## Proving Options


scripts/utils/bin/preflight.rs
Farhad-Shabani, line 1: Wonder shouldn't this binary be placed under the `scripts/utils/bin` directory?

book/fault_proofs/preflight.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Prerequisites
+
+- Deployed `DisputeGameFactory` contract
+- L1/L2 RPC access
+- SP1 network prover access
+
+## Required Environment Variables
+
+Create a `.env` file with the following variables:
```
Farhad-Shabani: Where should this live? inside the `/fault-proof` directory?
Also, if there’s already a `.env` file there from a previous e.g. deployment, would that cause any issues?
I’m thinking does it make sense to name it `.env.preflight`?

book/fault_proofs/preflight.md
```diff
@@ -0,0 +1,89 @@
...
+GAME_TYPE=42
+```
+
+### Network Configuration
+```bash
+# L1 RPC endpoint (used for Anvil fork during validation)
+L1_RPC=<YOUR_L1_RPC>
+
+# L1 beacon chain RPC endpoint
+L1_BEACON_RPC=<YOUR_L1_BEACON_RPC>
```
Farhad-Shabani: This might not be a necessary configuration. Right?

scripts/utils/bin/preflight.rs
```diff
@@ -0,0 +1,332 @@
...
+    info!("Agg proof saved to {agg_proof_dir}/agg.bin");
+
+    // 4. Spin up anvil.
+    let l1_head_number =
+        data_fetcher.l1_provider.get_block_by_hash(boot_info.l1Head).await?.unwrap().header.number;
+
+    let anvil = Anvil::new()
+        .fork(env::var("L1_RPC").expect("L1_RPC must be set"))
+        .fork_block_number(l1_head_number)
+        .args(["--no-mining"]);
```
Farhad-Shabani, line 308: Cool feature!

scripts/utils/bin/preflight.rs
fakedev9999: @Farhad-Shabani Good point. Is your comment coming from the idea that we only keep multi.rs and agg.rs in `scripts/prove/bin`?

Initially, I was thinking to place this under `fault-proof` dir because it is tightly coupled with fault proof infrastructure.

Happy to fix either way, which do you think makes more sense?

book/fault_proofs/preflight.md
fakedev9999: @Farhad-Shabani Yes, however it may be required if the L2 blocks being validated use blob data (EIP-4844).

Should I:
  1. Mark it as optional with a note about when it's needed?
  2. Keep it required to be safe for all scenarios?

book/fault_proofs/preflight.md
fakedev9999: Updated! Now defaults to `.env.preflight` to avoid conflicts.

Also updated documentation so that it now clearly states that `.env.preflight` should live in repository root and should be run from repository root and can use custom env file with --env-file flag.

See 8f81532.

book/fault_proofs/preflight.md
```diff
@@ -0,0 +1,97 @@
...
+# Set to 'false' to use a local private key (requires NETWORK_PRIVATE_KEY below).
+# Default: false
+USE_KMS_REQUESTER=false
+
+# SP1 network prover private key (required when USE_KMS_REQUESTER=false)
+# When USE_KMS_REQUESTER=true, this should be an AWS KMS key ARN instead
+NETWORK_PRIVATE_KEY=0x...
+```
+
+## Running Pre-Flight Validation
```
fakedev9999: Add feature flag for alt da

scripts/utils/bin/preflight.rs
Farhad-Shabani: > Is your comment coming from the idea that we only keep multi.rs and agg.rs in scripts/prove/bin?

As chatted, let's only keep `multi.rs` and `agg.rs` in `scripts/prove/bin` and place that under the `scripts/utils/bin` directory for now, alongside the others. 
Later, we should pull these out of the `/scripts` folder and figure out where they really belong.

book/fault_proofs/preflight.md
Farhad-Shabani: > Mark it as optional with a note about when it's needed?

As discussed, let's go with (1) to stay consistent with everywhere.

fakedev9999: @Farhad-Shabani Just figured out that this is not optional when proving.

See https://github.com/op-rs/kona/blob/15dac0759081de6683f900b4c20226eb8fa3b806/bin/host/src/single/cfg.rs#L283

Farhad-Shabani: Ahh, makes sense.

scripts/utils/bin/preflight.rs
```diff
@@ -0,0 +1,361 @@
...
+    provider: Arc<impl Provider + 'static>,
+    game_type: u32,
+) -> Result<u64> {
+    let factory = DisputeGameFactory::new(factory_address, provider.clone());
+
+    // Get the latest block number
+    let latest_block = provider.get_block_number().await?;
+
+    // Query in chunks to avoid exceeding max block range (100,000 blocks)
+    const CHUNK_SIZE: u64 = 100_000;
```
kien-rise: Let's make it configurable. The RPC endpoint I'm using does not allow me to exceed 10,000. Here is the code I am using:

```rust
let chunk_size: u64 = env::var("CHUNK_SIZE")
        .expect("CHUNK_SIZE must be set")
        .parse()
        .expect("CHUNK_SIZE must be a valid u64");
```

scripts/utils/bin/preflight.rs
```diff
@@ -0,0 +1,361 @@
...
+        parse_fulfillment_strategy(env::var("AGG_PROOF_STRATEGY")?),
+    )
+    .context("failed to determine network mode from range and agg fulfillment strategies")?;
+    let network_prover =
+        ProverClient::builder().network_for(network_mode).signer(network_signer).build();
+    info!("Initialized network prover successfully");
+
+    let (range_pk, _range_vk) = network_prover.setup(get_range_elf_embedded());
+    let mut range_proof =
+        network_prover.prove(&range_pk, &range_proof_stdin).compressed().run().unwrap();
```
kien-rise: this call needs `.strategy(parse_fulfillment_strategy(env::var("RANGE_PROOF_STRATEGY")?))`

scripts/utils/bin/preflight.rs
```diff
@@ -0,0 +1,361 @@
...
+                "Invalid AGG_PROOF_MODE '{}'. Expected one of: plonk, groth16",
+                other
+            ))
+        }
+    };
+    info!("Aggregation proof mode: {:?}", agg_proof_mode);
+
+    let (agg_pk, _) = network_prover.setup(AGGREGATION_ELF);
+    let agg_proof =
+        network_prover.prove(&agg_pk, &agg_proof_stdin).mode(agg_proof_mode).run().unwrap();
```
kien-rise: this call needs `.strategy(parse_fulfillment_strategy(env::var("AGG_PROOF_STRATEGY")?))`

PR #670, opened 2025-10-30
refactor(scripts): default to network prover

fakedev9999:
Defaults to network prover for range proof and agg proof scripts when `--prove` flag is provided.

This is to avoid confusion around the EnvProver, which falls to CpuProver by default which runs for a very long time even for a small range proof.

### DIFF
--- a/scripts/prove/bin/agg.rs
+++ b/scripts/prove/bin/agg.rs
@@ -103,15 +103,14 @@ async fn main() -> Result<()> {
     println!("Aggregate ELF Verification Key: {:?}", agg_vk.vk.bytes32());
 
     if args.prove {
-        let agg_proof_mode = if env::var("AGG_PROOF_MODE")
-            .unwrap_or_else(|_| "plonk".to_string())
-            .to_lowercase()
-            == "groth16"
-        {
-            SP1ProofMode::Groth16
-        } else {
-            SP1ProofMode::Plonk
-        };
+        let agg_proof_mode =
+            if env::var("AGG_PROOF_MODE").unwrap_or_else(|_| "plonk".to_string()).to_lowercase() ==
+                "groth16"
+            {
+                SP1ProofMode::Groth16
+            } else {
+                SP1ProofMode::Plonk
+            };
 
         prover
             .prove(&agg_pk, &stdin)


scripts/prove/bin/agg.rs
```diff
@@ -72,7 +83,7 @@ async fn main() -> Result<()> {
 
     dotenv::from_filename(args.env_file).ok();
 
-    let prover = ProverClient::from_env();
+    let prover = ProverClient::builder().network().build();
```
Farhad-Shabani, line 86: I’m thinking we should align how we build the prover with how we already do it for proposers, similar to here:
https://github.com/succinctlabs/op-succinct/blob/7714073965acbc84b43647f8007dbbcf0be6caf8/fault-proof/src/proposer.rs#L200-L204
So can shrink the code surface and reuse the same construction logic. Does that make sense?

scripts/prove/bin/multi.rs
```diff
@@ -46,13 +54,20 @@ async fn main() -> Result<()> {
     // Get the stdin for the block.
     let sp1_stdin = host.witness_generator().get_sp1_stdin(witness_data)?;
 
-    let prover = ProverClient::from_env();
-
     if args.prove {
         // If the prove flag is set, generate a proof.
+        let prover = ProverClient::builder().network().build();
```
Farhad-Shabani, line 59: Same comment as: https://github.com/succinctlabs/op-succinct/pull/670/files#r2561346960

book/advanced/prove-scripts.md
```diff
@@ -0,0 +1,174 @@
...
+## Generating Range Proofs
+
+The `multi.rs` binary generates compressed range proofs for a specified block range. These proofs verify the state transition function for the L2 blocks in the range.
+
+### Usage
+
+```bash
+cargo run --bin multi --release -- \
+    --start <START_BLOCK> \
+    --end <END_BLOCK> \
```
Farhad-Shabani, line 95: Based on what we found yesterday, `--batch-size` should always match `end - start` for the command to work correctly. When both end and start are provided, we should default the batch size to that value. We can handle that in a separate PR.

book/advanced/prove-scripts.md
```diff
@@ -0,0 +1,174 @@
+# Prove Scripts
```
Farhad-Shabani, line 1: 👌🏻

fakedev9999:
Closing — this is fully superseded by #820, which landed on main on 2026-03-06.

#820 covers everything in this PR plus Farhad's review feedback:
- Extracted `build_network_prover_from_env(strategy)` shared helper (reuses `get_network_signer` + `network_for(mode)` pattern)
- `multi` and `agg` default to network prover with `RANGE_PROOF_STRATEGY` / `AGG_PROOF_STRATEGY` defaulting to `"reserved"`
- Execute-only path uses `CpuProver` so `NETWORK_PRIVATE_KEY` is only required with `--prove`
- Cluster mode integration from #805
- Prove scripts docs consolidated into `book/advanced/prove-scripts.md`

@seolaoh The Celo use case (`RANGE_PROOF_STRATEGY=reserved` on `multi`) is already on main — pulling main or the next release should cover it without cherry-picking this branch.

PR #672, opened 2025-11-01
docs: fix fault-proof proposer.md

fakedev9999:
Improves fault-proof proposer doc by:
1. fixing typo of the env file name
2. removing unnecessary direction for sourcing env file


Farhad-Shabani (changes requested):
Should we update `challenger.md`? That points to the `root` directory.
Actually, there are other places (search for `.env.proposer`), we direct readers to the `root` directory.

PR #673, opened 2025-11-02
chore: improve fault proof contracts upgrade flow

fakedev9999:
Improves fault proof contracts upgrade flow by:
1. removing the private key requirement for dry run
2. adding missing requirements in the doc.


Farhad-Shabani (approved):
👌🏻

PR #674, opened 2025-11-02
feat: add sp1Verifier() getter to OPSuccinctFaultDisputeGame

fakedev9999:
Adds public getter for the `SP1_VERIFIER` immutable variable to maintain consistency with other immutable getters and improve transparency.

This allows:
- Users/contracts to verify which SP1 verifier is being used
- Off-chain tools to validate proof verification configurations
- Easier debugging and deployment validation

Follows the existing pattern of exposing all immutable constructor parameters (disputeGameFactory, rollupConfigHash, aggregationVkey, etc.).


Farhad-Shabani (approved):
Makes sense!

PR #676, opened 2025-11-05
fix(fault-proof): retain anchor game in cache

fakedev9999:
Fixes the fault proof proposer to retain the anchor game in the cache.

This is problematic in cases when the anchor game is set after finalization and bond claim, the game is removed from the cache, leading proposer to create a new game from the anchor game, which results in an infinite loop with a GameAlreadyExists() error.

### DIFF
--- a/justfile
+++ b/justfile
@@ -358,7 +358,7 @@ remove-config config_name env_file=".env":
 
 # Run all unit and integration tests except for the specified ones.
 tests:
-   cargo t --release --workspace --exclude op-succinct-fp \
+   cargo t --release \
     -- \
     --skip test_cycle_count_diff \
     --skip test_post_to_github


fault-proof/src/proposer.rs
```diff
@@ -414,16 +444,38 @@ where
...
+                    // Game removal policy:
+                    // - Canonical head games are retained even with zero credit to maintain chain
+                    //   consistency.
+                    // - Anchor games are retained as they serve as the root of the dispute game
+                    //   tree.
+                    // - All other games with bonds already claimed are removed to free cache
+                    //   memory.
                     GameSyncAction::Remove(index) => {
-                        let is_canonical_head = state.canonical_head_index == Some(index);
+                        let Some(game) = state.games.get(&index) else {
```
Farhad-Shabani: The actions, just one step back, are derived from games. And those already inserted and exist in the cache. Logically never this fall into the `else`, but if ever (don’t know how?), that’s an error.

fault-proof/src/proposer.rs
```diff
@@ -414,16 +444,38 @@ where
                             game.should_attempt_to_claim_bond = should_attempt_to_claim_bond;
                         }
                     }
+                    // Game removal policy:
+                    // - Canonical head games are retained even with zero credit to maintain chain
+                    //   consistency.
+                    // - Anchor games are retained as they serve as the root of the dispute game
+                    //   tree.
+                    // - All other games with bonds already claimed are removed to free cache
+                    //   memory.
```
Farhad-Shabani: I was thinking about this a bit more. I think this is not the right spot to enforce the policy. It makes more sense a few lines up, where we decide which action gets added. Logically, we shouldn’t even be appending a `Remove` action for those cases in the first place. WDYT?

justfile
```diff
@@ -358,10 +358,14 @@ remove-config config_name env_file=".env":
...
-   cargo t --release \
+   cargo t --release --workspace --exclude op-succinct-fp \
     -- \
     --skip test_cycle_count_diff \
-    --skip test_post_to_github \
+    --skip test_post_to_github
+
+# Run fault-proof integration tests.
+fp-integration-tests:
+   cargo t --release -p op-succinct-fp --test integration_test -- --test-threads=1 --nocapture
```
Farhad-Shabani: I honestly don’t see much difference between this test and the others under e2e. They all feel more like integration tests since several components are mocked or missed. Let’s keep it under `e2e` for now and rename/reorg them all in a separate PR later.

fault-proof/src/proposer.rs
```diff
@@ -138,6 +138,14 @@ impl ProposerState {
     }
 }
 
+/// Snapshot of the proposer's cached state for testing and monitoring.
+#[derive(Clone, Debug)]
+pub struct ProposerStateSnapshot {
+    pub anchor_index: Option<U256>,
```
Farhad-Shabani, line 144: This is exactly what I mentioned today. The system only cares about the anchor index!

fault-proof/src/proposer.rs
fakedev9999: Were you expecting something like
```
let game = state.games.get(&index).with_context(|| {
    format!(
        "Game {index} scheduled for removal but missing from proposer cache"
    )
})?;
```
?

So do you think fine it's okay to revert the entire actions for loop (`for action in actions`) since it should never happen?

My initial purpose was to prevent cache from remaining stale until the next tick because we have other operations (game creation, resolution, bond claim) scheduled based on the cache state after this function returns.

fault-proof/src/proposer.rs
fakedev9999: Yeah I found that too! Do you want me to include the refacor in this PR? Or should we keep it in a separate ticket?

fault-proof/src/proposer.rs
fakedev9999: That's actually so true.

fakedev9999: 1542455

fault-proof/src/proposer.rs
Farhad-Shabani: Oh, didn't talked well myself. I just meant to change the logging level to `error` level. 
But I see with the new changes this is gone. cool! 🙂

fault-proof/src/proposer.rs
Farhad-Shabani: Yep, let's handle that separately!

justfile
```diff
@@ -358,10 +358,10 @@ remove-config config_name env_file=".env":
 
 # Run all unit and integration tests except for the specified ones.
 tests:
-   cargo t --release \
+   cargo t --release --workspace --exclude op-succinct-fp \
```
Farhad-Shabani: Should revert this?

Farhad-Shabani (approved):
Just a small nit! All set otherwise. Thanks!

PR #677, opened 2025-11-05
fix(fault-proof): skip legacy games

Farhad-Shabani:
Closes: GRO2-878

The FP proposer wasn’t handling game-type transitions, and the existing test had a bug that incorrectly made it look like it was.

With this change, the proposer now skips any unsupported game types when storing to cache. The `anchor_game` remains `None` until we have an anchor with a supported game type. That’s OK, because on startup the proposer [queries the anchor L2 block number](https://github.com/succinctlabs/op-succinct/blob/b40cafe1eb45dcf1120b1c0e63c24dca75605130/fault-proof/src/proposer.rs#L191) and starts creating games from there.

- If the transition has already happened, it just sees legacy game types, skips them, and starts deploying games from the anchor L2 block.
- If the transition hasn’t happened yet, game creation fails and keeps retrying until the transition occurs.

In that latter case we should eventually add more graceful handling and better logging, but that’s out of scope for this PR.

Also this PR:

- Fixes the faulty `test_game_type_transition_skips_legacy_games`.
- Adds `test_game_type_transition_while_proposer_running` to ensure the proposer keeps running and resolving games as expected during the transition.
- Improves the loggings around

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -837,7 +837,7 @@ where
     ///
     /// Drop game if:
     /// - The game type is not supported.
-    /// - The game type deos not respect the expected type when created.
+    /// - The game type does not respect the expected type when created.
     /// - The output root claim is invalid.
     /// - The parent game does not exist in cache if it should have one.
     async fn fetch_game(&self, index: U256) -> Result<()> {
--- a/fault-proof/tests/e2e.rs
+++ b/fault-proof/tests/e2e.rs
@@ -423,9 +423,9 @@ mod e2e {
         .await?;
         info!("✓ Respected game type restored to {TEST_GAME_TYPE}");
 
-        let factory = DisputeGameFactory::new(env.deployed.factory, env.anvil.provider.clone());
         let tracked_games =
-            wait_and_track_games(&factory, TEST_GAME_TYPE, 3, Duration::from_secs(120)).await?;
+            wait_and_track_games(&factory_reader, TEST_GAME_TYPE, 3, Duration::from_secs(120))
+                .await?;
         assert_eq!(tracked_games.len(), 3);
         info!("✓ Proposer created 3 type {} games despite legacy history", TEST_GAME_TYPE);
 


fault-proof/tests/e2e.rs
```diff
@@ -200,55 +204,44 @@ mod e2e {
         .await?;
 
         let mut expected_index = initial_game_count;
-        let mut mock_game_addresses = Vec::with_capacity(MOCK_PERMISSIONED_GAMES_TO_SEED);
+        info!("Seeding legacy game (type {MOCK_PERMISSIONED_GAME_TYPE})");
 
-        info!(
-        "Seeding {MOCK_PERMISSIONED_GAMES_TO_SEED} legacy games (type {MOCK_PERMISSIONED_GAME_TYPE})"
-    );
-        for i in 0..MOCK_PERMISSIONED_GAMES_TO_SEED {
```
Farhad-Shabani, line 208: Changed it to deploy just one mock legacy game instead of three. In practice, it doesn’t add any extra coverage. Both setups behave the same. Hence simplified. (unless missing something?)

contracts/src/utils/MockPermissionedDisputeGame.sol
```diff
@@ -1,23 +1,50 @@
...
     Timestamp private _resolvedAt;
+    ClaimData public claimData;
+    OutputRoot private _startingOutputRoot;
+
+    constructor(IDisputeGameFactory _disputeGameFactory) {
+        DISPUTE_GAME_FACTORY = _disputeGameFactory;
+    }
 
     function initialize() external payable override {
         require(!initialized, "already initialized");
```
fakedev9999, line 31:
```
// INVARIANT: The game can only be initialized by the dispute game factory.
        if (address(DISPUTE_GAME_FACTORY) != msg.sender) revert IncorrectDisputeGameFactory();
```
nit: something like above like the actual fault dispute game?

(I guess it's fine not to have it since it's mock contract. You can resolve it if you don't think it's necessary since it's mock.)

fault-proof/src/proposer.rs
```diff
@@ -835,48 +835,98 @@ where
 
     /// Fetch game from the factory.
     ///
-    /// Drop game if the game type is invalid or the output root is not valid.
-    /// Drop game if the parent game does not exist.
+    /// Drop game if:
+    /// - The game type is not supported.
+    /// - The game type deos not respect the expected type when created.
```
```suggestion
    /// - The game type does not respect the expected type when created.
```

fault-proof/src/proposer.rs
```diff
@@ -835,48 +835,98 @@ where
...
+    /// - The output root claim is invalid.
+    /// - The parent game does not exist in cache if it should have one.
     async fn fetch_game(&self, index: U256) -> Result<()> {
+        let mut state = self.state.lock().await;
```
fakedev9999, line 844: I'm pretty sure it's fine to hold lock here since we fetch games sequentially and all other tasks await for `sync_games`.
But just wanted to cross check with you since the best practice is to "lock only when we need to touch state".

contracts/src/utils/MockPermissionedDisputeGame.sol
```diff
@@ -39,42 +66,41 @@ contract MockPermissionedDisputeGame is IDisputeGame {
...
+        gameType_ = gameType();
+        rootClaim_ = rootClaim();
+        extraData_ = extraData();
     }
 
-    /// @notice Legacy games expose `claimData(uint256)` rather than zero-arg version.
-    function claimData(uint256) external pure returns (bytes memory) {
-        return bytes("");
+    function wasRespectedGameTypeWhenCreated() external pure returns (bool) {
+        return true;
```
fakedev9999, line 104: Shouldn't we actually do something like the actual fault dispute game?
```
wasRespectedGameTypeWhenCreated =
    GameType.unwrap(ANCHOR_STATE_REGISTRY.respectedGameType()) == GameType.unwrap(GAME_TYPE);
```
.

Then I think we should bake the anchor state registry into the mock dispute game as well.

fault-proof/tests/e2e.rs
```diff
@@ -293,20 +287,178 @@ mod e2e {
...
+            _impl: legacy_impl,
+        };
+        send_contract_transaction(
+            &proposer_signer,
+            &l1_rpc_url,
+            env.deployed.factory,
+            Bytes::from(set_impl_call.abi_encode()),
+            None,
+        )
+        .await?;
```
fakedev9999, line 352: Could we make this a common test helper function? I think the setup is quite similar to `test_game_type_transition_skips_legacy_game` test.

fault-proof/src/proposer.rs
```diff
@@ -835,48 +835,98 @@ where
...
-            Err(error) => {
-                tracing::debug!(game_index = %index, ?game_address, ?error,
-                    "Falling back to legacy game with dummy claim data");
-                (u32::MAX, ProposalStatus::Unchallenged, 0)
-            }
-        };
-
         let was_respected = contract.wasRespectedGameTypeWhenCreated().call().await?;
         let status = contract.status().call().await?;
+        let claim_data = contract.claimData().call().await?;
```
fakedev9999, line 870: What do you think of panicking here since we filter foreign game types above?

fault-proof/tests/e2e.rs
```diff
@@ -293,20 +287,178 @@ mod e2e {
...
+            &proposer_signer,
+            &l1_rpc_url,
+            env.deployed.portal,
+            Bytes::from(restore_type_call.abi_encode()),
+            None,
+        )
+        .await?;
+        info!("✓ Respected game type restored to {TEST_GAME_TYPE}");
+
+        let factory = DisputeGameFactory::new(env.deployed.factory, env.anvil.provider.clone());
```
fakedev9999:
```
        let factory_reader =
            DisputeGameFactory::new(env.deployed.factory, env.anvil.provider.clone());
```
nit: I think we can reuse the factory initialized above.

fault-proof/tests/e2e.rs
```diff
@@ -293,20 +287,178 @@ mod e2e {
...
+            GameStatus::IN_PROGRESS,
+            "mock permissioned game should remain untouched"
+        );
+
+        Ok(())
+    }
+
+    // Ensures the proposer can handle a game type transition while running.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_game_type_transition_while_proposer_running() -> Result<()> {
```
fakedev9999, line 307: After seeing this test, I think we should add a condition to the proposer to prevent game creation when the configured game type is not respected game type.

From the test logs, I see the proposer is actually trying to create games.

I also think there should be also a check before the respected game type update to the portal which checks that game count didn't increase.

contracts/src/utils/MockPermissionedDisputeGame.sol
Farhad-Shabani: Yeah, let’s keep it that way. It gives us flexibility later if just want to run a unit or integration test w/o setting up through the factory.

fault-proof/tests/e2e.rs
Farhad-Shabani: I did intend that. The problem is that the shared function would need to return several components like `env`, `init_bond`, `proposer_signe`, `factory_reader`, and some of those would still have to be reconstructed from env afterward.
To keep things cleaner, we should instead make a separate change to extend `TestEnvironment` with helper methods. That way, tests can access only what they need without manually wiring all the pieces together.
If you agree, let’s handle that in a separate PR?

fault-proof/tests/e2e.rs
Farhad-Shabani: Makes sense. I opened a ticket for this. See `GRO2-890` please.

> I also think there should be also a check before the respected game type update to the portal which checks that game count didn't increase.

Not sure I got you correctly 🤔

fault-proof/src/proposer.rs
Farhad-Shabani: What do you mean by “panic” here? Earlier, we were worried about unsupported game types, but since we now filter out foreign ones above, that shouldn’t be an issue anymore. Wouldn’t this just behave like the other calls now?

contracts/src/utils/MockPermissionedDisputeGame.sol
Farhad-Shabani: Good point. I actually thought about that. I don’t think it’s necessary, honestly. The thing is, since the factory is the creator, we can assume the mock games are already respected in our tests. So adding more would just bloat the mock.

fault-proof/tests/e2e.rs
Farhad-Shabani: [4e4b972](https://github.com/succinctlabs/op-succinct/pull/677/commits/4e4b972b1ab03c117f95b78726531a24482d77e7)

contracts/src/utils/MockPermissionedDisputeGame.sol
fakedev9999: Yeah let's just keep it simple.

fault-proof/tests/e2e.rs
fakedev9999: Sure can you open a ticket for that and resolve the comment?

fault-proof/src/proposer.rs
fakedev9999: I was meaning something like `let claim_data = contract.claimData().call().await.expect()`.
But now that I'm thinking, it's better to keep it as is. It's not good to crash here. We should handle errors gracefully.

fault-proof/tests/e2e.rs
fakedev9999: Left some comments in the Linear ticket, let's discuss further in that thread.

fault-proof/src/proposer.rs
Farhad-Shabani: [Looks like my comment from yesterday disappeared somehow]

For now, I just sticked to the existing pattern. Called it here because we now need to set the cursor when the game type isn’t supported. You’re totally right though, that tripped me up too. The better fix would be to lock right before touch and unlock right after. That would mean `OPSuccinctProposer` should expose a few methods to handle that internally, and we’d just call like `cursor.set(index)` here. That’s more of a separate PR though.

fault-proof/tests/e2e.rs
Farhad-Shabani: Of course. Opened `GRO2-902`

fault-proof/src/proposer.rs
Farhad-Shabani: Opened: `GRO2-904`

PR #679, opened 2025-11-06
fix(fault-proof): derive proving interval from contract

Farhad-Shabani:
Closes: GRO2-891
Closes: GRO2-887

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -1418,8 +1418,7 @@ where
         let game = OPSuccinctFaultDisputeGame::new(game_address, self.l1_provider.clone());
         let starting_l2_block_number = game.startingBlockNumber().call().await?;
         let l2_block_number = game.l2BlockNumber().call().await?;
-        let start_block =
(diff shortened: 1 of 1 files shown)


fault-proof/src/proposer.rs
```diff
@@ -1417,8 +1416,10 @@ where
 
         // Get the game block number to include in logs
         let game = OPSuccinctFaultDisputeGame::new(game_address, self.l1_provider.clone());
+        let starting_l2_block_number = game.startingBlockNumber().call().await?;
         let l2_block_number = game.l2BlockNumber().call().await?;
-        let start_block = l2_block_number.to::<u64>() - self.config.proposal_interval_in_blocks;
+        let start_block =
+            l2_block_number.to::<u64>().saturating_sub(starting_l2_block_number.to::<u64>());
```
fakedev9999: I think this code is bugged. This calculates the difference not the actual starting block.

We should just do
```suggestion
        let start_block = game.startingBlockNumber().call().await?;
```

Farhad-Shabani: Ohh, right. Nice catch!

PR #681, opened 2025-11-08
fix: avoid full replay on restarts by backward sync

Farhad-Shabani:
Closes: `GRO2-892`

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -100,18 +100,14 @@ pub struct Game {
 /// Tracks:
 /// - `anchor_game`: the latest anchor fetched from the registry
 /// - `canonical_head_index`/`canonical_head_l2_block`: the best known game for scheduling work
-/// - `start_cursor`: the index in the factory's dispute-game list to start from during incremental
-///   reverse-chronological syncs
-/// - `end_cursor`: the last index in the factory's dispute-game list processed during incremental
-///   backward syncs
+/// - `cursor`: the last index of factory's dispute game list processed during incremental syncs
 /// - `games`: cached metadata for every tracked game keyed by index
 #[derive(Default)]
 struct ProposerState {
     anchor_game: Option<Game>,
     canonical_head_index: Option<U256>,
     canonical_head_l2_block: Option<U256>,
-    start_cursor: Option<U256>,
-    end_cursor: U256,
+    cursor: U256,
     games: HashMap<U256, Game>,
 }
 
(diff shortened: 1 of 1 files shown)


fault-proof/src/proposer.rs
```diff
@@ -247,6 +256,7 @@ where
             // 1. Synchronize cached dispute state before scheduling work.
             if let Err(e) = self.sync_state().await {
                 tracing::warn!("Failed to sync proposer state: {:?}", e);
+                continue
```
Farhad-Shabani, line 259: ⁉️

fakedev9999: Makes sense to do no operations when the cache is stale!

fault-proof/src/proposer.rs
```diff
@@ -297,21 +310,56 @@ where
...
+                        if anchor_d.abs_diff(deadline) > MAX_GAME_DEADLINE_LAG {
+                            break;
+                        }
+                    }
+                }
+            }
+
+            if index == cursor {
+                break;
+            }
```
fakedev9999: Given that it starts the loop at latest_index and only exits when index == cursor, I think we should add a guard for the case where cursor is larger than the on-chain latest.

This should basically never/rarely happen but I can imagine a case where the factory is redeployed/reset while the proposer keeps running.

In such situation, the index quickly saturates to 0, never equals cursor, none of the other break conditions trigger, and sync_games will never end.

fault-proof/src/proposer.rs
```diff
@@ -297,21 +310,56 @@ where
...
+
+            index = index.saturating_sub(U256::from(1));
+        }
+
+        {
+            let mut state = self.state.lock().await;
+            state.cursor = latest_index;
         }
```
fakedev9999, line 378: Now cursor is only updated after the backward scan finishes, so if any fetch_game call fails mid-loop (RPC timeout, bad response, etc.), the function returns with the cursor still pointing at the previous "stable" value. 

The next tick will re-fetch every index that already succeeded, and if one troublesome index keeps erroring the proposer will redo the entire suffix on every poll and never make forward progress.

Can we consider some bookkeeping method in the sync so that  we don’t replay the entire suffix after a hiccup?

fault-proof/src/proposer.rs
Farhad-Shabani: See if [372e9e7](https://github.com/succinctlabs/op-succinct/pull/681/commits/372e9e739e40e91f1a1cf5051446dbc815c7a647) works.

fakedev9999: @Farhad-Shabani Wdyt of having a warning log for better visibility?

fault-proof/src/proposer.rs
Farhad-Shabani: Makes sense! Good call.
Here's the fix: [8412762](https://github.com/succinctlabs/op-succinct/pull/681/commits/84127627bfc2d31b6519b0e6c498603b0296e9c0)

fault-proof/src/proposer.rs
Farhad-Shabani: [0b6981d](https://github.com/succinctlabs/op-succinct/pull/681/commits/0b6981d8c6198b6705bd5c5e49c0c8569e17919e)

fault-proof/src/proposer.rs
Farhad-Shabani: Ah, just realized, the fix doesn’t always work as expected.

Farhad-Shabani: Fixed another way: [c0001c1](https://github.com/succinctlabs/op-succinct/pull/681/commits/c0001c109f3ec8225ab1e5da1d3c665993b3b41d)

fakedev9999 (approved):
lgtm

PR #682, opened 2025-11-08
refactor: filter e2e logging targets

Farhad-Shabani:
Closes: `GRO2-911`

Just tossed in the logging targets we actually care about.  Debugging should be easier now!

### DIFF
--- a/fault-proof/tests/common/env.rs
+++ b/fault-proof/tests/common/env.rs
@@ -104,7 +104,7 @@ pub fn init_logging() {
             ("e2e", level),
             ("fault_proof", level),
             ("op_succinct_fp", level),
-            ("op-succinct-client-utils", level),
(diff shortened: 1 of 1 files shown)


fault-proof/tests/common/env.rs
```diff
@@ -96,3 +91,26 @@ impl TestEnvironment {
...
+
+    INIT.get_or_init(|| {
+        let level =
+            std::env::var("RUST_LOG").unwrap_or("info".to_string()).parse().unwrap_or(Level::INFO);
+
+        let filter = Targets::new().with_targets([
+            ("e2e", level),
+            ("fault_proof", level),
+            ("op_succinct_fp", level),
+            ("op-succinct-client-utils", level),
```
fakedev9999:
```suggestion
            ("op_succinct_client_utils", level),
```

Issue #665, opened 2025-10-27
Review and fix actions triggers


Farhad-Shabani:
Right now, most (and sometimes all) CI jobs run on every PR, even when the changes don’t need a full pipeline, like docs-only updates. We should review the workflow triggers and make sure only the right jobs run for each type of change.

PR #684, opened 2025-11-11
fix: adjust action triggers + concurrency

Farhad-Shabani:
Closes: `GRO2-914`
Closes: #665

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,6 +2,8 @@ name: Cargo Tests
 
 on:
   pull_request:
+    branches:
+      - main
     paths:
       - '.github/workflows/ci.yml'
       - 'Cargo.toml'
@@ -10,16 +12,21 @@ on:
       - 'fault-proof/**'
       - 'bindings/**'
       - 'configs/**'
-      - 'elfs/**'
+      - 'elf/**'
       - 'programs/**'
       - 'scripts/**'
       - 'utils/**'
       - 'validity/**'
-    tags:
-      - v[0-9]+.*
+
+  push:
     branches:
-      - "release/*"
       - main
+      - "release/*"
+    tags:
+      - "v[0-9]+.*"
+    paths-ignore:
+      - "**/*.md"
+      - "**/*.pdf"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -2,18 +2,36 @@ name: Cargo Tests
...
+      - 'fault-proof/**'
+      - 'bindings/**'
+      - 'configs/**'
+      - 'elfs/**'
+      - 'programs/**'
+      - 'scripts/**'
+      - 'utils/**'
+      - 'validity/**'
+    tags:
+      - v[0-9]+.*
```
fakedev9999: This won't be triggered since GitHub doesn’t allow tag filters on PR triggers. 

Let's wire them up under a push block instead like other workflows.

.github/workflows/ci.yaml
```diff
@@ -2,18 +2,36 @@ name: Cargo Tests
...
   pull_request:
+    paths:
+      - '.github/workflows/ci.yml'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'justfile'
+      - 'fault-proof/**'
+      - 'bindings/**'
+      - 'configs/**'
+      - 'elfs/**'
```
fakedev9999:
```suggestion
      - 'elf/**'
```

.github/workflows/ci.yaml
```diff
@@ -2,18 +2,36 @@ name: Cargo Tests
 
 on:
   pull_request:
+    paths:
```
fakedev9999, line 7: This paths filter adds maintenance cost. Every time you rename folders or add new areas that should trigger CI, someone has to remember to update the workflow or it silently doesn't run.

.github/workflows/ci.yaml
Farhad-Shabani: Ah yep, I mistakenly dropped that under `pull_requests`. Fixed.

.github/workflows/ci.yaml
Farhad-Shabani: Fair, I see where you’re coming from, but I’ll have to disagree 🙂
Running CI on irrelevant changes adds more long-term cost: slower dev cycles and higher CI spend. Particularly as test coverage grows, keeping CI runs focused will matter even more.

We’re unlikely to rename folders much now. When we do, I think it's easy to update paths. 
Actually, regex-like filters could be a better option, but not worth to explore that right now.

PR #687, opened 2025-11-11
ci: Run e2e tests when dependencies change

piersy:
Almost all the dependencies in fault-proof and bindings reference
workspace for their versions, so if the workspace versions change
the e2e tests should be run.


Farhad-Shabani:
Thanks for flagging this! This is handled as part of #684.

PR #689, opened 2025-11-12
fix: fetch and use the correct l2_chain_id

kien-rise:
The L2 chain id is used here:

https://github.com/op-rs/kona/blob/kona-node/v1.2.2/crates/proof/proof/src/boot.rs

The current code uses `None`, which is later converted to `0` by `.unwrap_or_default()`. By passing the wrong chain id, Kona code will use the fallback logic (which is still fine but not optimal).

### DIFF
--- a/utils/host/src/fetcher.rs
+++ b/utils/host/src/fetcher.rs
@@ -677,10 +677,9 @@ impl OPSuccinctDataFetcher {
         l2_end_block: u64,
         l1_head_hash: B256,
     ) -> Result<SingleChainHost> {
-        // If the rollup config is not already loaded, fetch and save it.
-        if self.rollup_config.is_none() {
+        let Some(rollup_config) = &self.rollup_config else {
             return Err(anyhow::anyhow!("Rollup config not loaded."));
-        }
+        };
 
         if l2_start_block >= l2_end_block {
             return Err(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


kien-rise:
Without this , I see warning:

```
WARN [[Proving]]{start_block=21777043 end_block=21777053 game_address=0x0915906934a9f437193a95114f3b906032779ba2}: No rollup config found for chain ID 0, falling back to preimage oracle. This is insecure in production without additional validation!
```

utils/host/src/fetcher.rs
```diff
@@ -743,7 +744,7 @@ impl OPSuccinctDataFetcher {
             agreed_l2_head_hash,
             claimed_l2_output_root,
             claimed_l2_block_number: l2_end_block,
-            l2_chain_id: None,
+            l2_chain_id: Some(l2_chain_id),
```
fakedev9999:
```suggestion
            l2_chain_id: Some(self.rollup_config.as_ref().unwrap().l2_chain_id),
```
Nice improvement! One minor suggestion:

The `rollup_config` is already loaded and validated at this point (line 681 checks `self.rollup_config.is_none()`). We can avoid the extra RPC call by using the cached value:

kien-rise: nice! I have applied your suggestion.

PR #690, opened 2025-11-13
chore: bump to kona 1.1.7, hokulea 1.0.2 and v3.3.3b

fakedev9999:
This PR is opened to review the diff between v3.3.2 and v3.3.3b prior to the patch release.


fakedev9999:
@Farhad-Shabani Was just waiting for Celo to confirm they're on op-node lower than v1.14.3.

PR #691, opened 2025-11-13
fix: use `RwLock` for state + lower `fetch_interval` for tests

Farhad-Shabani:


fault-proof/tests/e2e.rs
```diff
@@ -907,20 +888,7 @@ mod e2e {
...
-                parent_index: game.parent_index,
-                l2_block: game.l2_block_number,
-                status: GameStatus::DEFENDER_WINS,
-                proposal_status: ProposalStatus::Resolved,
-                deadline: 0,
-                should_attempt_to_resolve: false,
-                should_attempt_to_claim_bond: true,
-            };
-            proposer.submit_bond_claim_transaction(&game).await?;
-        }
```
Farhad-Shabani, line 923: Because this relies on a secondary proposer, timing becomes a factor, and that isn’t part of this test case. This was preventing to shorten the fetch interval. Once the interval is shorter, the active proposer itself was handling resolution and claiming. Changed now to only rely on the running proposer.

fault-proof/src/proposer.rs
```diff
@@ -175,7 +178,7 @@ where
     host: Arc<H>,
     tasks: Arc<Mutex<TaskMap>>,
     next_task_id: Arc<AtomicU64>,
-    state: Arc<Mutex<ProposerState>>,
+    state: Arc<RwLock<ProposerState>>,
```
Farhad-Shabani, line 181: Before we were on mutex-lock state even for reads, so when tests try to snapshot state while the proposer is running, they sometimes can’t get access and end up hanging. Switched it to `tokio::sync::RwLock` so can drop the `fetch_interval` further.

fault-proof/tests/e2e.rs
```diff
@@ -930,22 +898,30 @@ mod e2e {
...
+
+        // Wait for proposer to create a new game beyond the invalidated canonical head
+        let mut i = U256::from(FIRST_NEW_GAME_INDEX);
+        while current_game_count <= i {
+            tokio::time::sleep(Duration::from_secs(1)).await;
             current_game_count = factory.gameCount().call().await?;
         }
 
-        // Verify the new game is built on the last valid game at index 1
-        let new_game_index = U256::from(3);
```
Farhad-Shabani, line 939: When lower the interval, the `new_game_index` right above the last valid one isn’t guaranteed to be 3. Since more games can be created in the meantime, there may be several additional games stacked on top of that invalidated one.

fault-proof/src/proposer.rs
```diff
@@ -390,19 +393,19 @@ where
...
+            let mut state = self.state.write().await;
             state.cursor = latest_index;
         }
 
         for idx in invalid_game_ids {
             tracing::warn!(game_index = %idx, "Removing invalid game and its subtree from cache");
-            let mut state = self.state.lock().await;
+            let mut state = self.state.write().await;
             state.remove_subtree(idx);
         }
```
```suggestion
        if !invalid_game_ids.is_empty() {
            let mut state = self.state.write().await;
            for idx in invalid_game_ids {
                tracing::warn!(
                    game_index = %idx,
                    "Removing invalid game and its subtree from cache"
                );
                state.remove_subtree(idx);
            }
        }
```

What do you think of changing it to hold a single write lock while removing all subtrees?

Farhad-Shabani: [dec29e9](https://github.com/succinctlabs/op-succinct/pull/691/commits/dec29e937367686702dc1f79b6b0442e7abd482e)

PR #694, opened 2025-11-14
test(fault-proof): add integrations test for unsupported/invalid games

fakedev9999:
Closes GRO2-951 with `test_invalid_game_type_handling` and `test_sync_state_filters_multiple_legacy_games`.
Removes redundant `test_game_type_transition_skips_legacy_game` and `test_game_type_transition_while_proposer_running` e2e test.

Closes GRO2-978 with `test_sync_state_filters_non_respected_game_type`.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -45,10 +45,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: peaceiris/actions-mdbook@v1
         with:
-          mdbook-version: 'latest'
+          mdbook-version: '0.4.52'
       - name: Install plugins
         run: |
-          cargo install mdbook-mermaid
+          cargo install mdbook-mermaid@0.16.0
           cargo install mdbook-admonish
       - name: Build book
         run: mdbook build
--- a/book/fault_proofs/docker.md
+++ b/book/fault_proofs/docker.md
@@ -96,7 +96,7 @@ docker compose down
 
 For fast finality mode, add these to your `.env.proposer`:
 
-```
+```bash
 FAST_FINALITY_MODE=true
 NETWORK_PRIVATE_KEY=     # Private key for the Succinct Prover Network
 L1_BEACON_RPC=           # L1 Beacon RPC endpoint URL
--- a/book/fault_proofs/proposer.md
+++ b/book/fault_proofs/proposer.md
@@ -78,6 +78,8 @@ Depending on the one you choose, you must provide the corresponding environment 
 
 | Variable | Description | Default Value |
 |----------|-------------|---------------|
+| `L1_CONFIG_DIR` | The directory containing the L1 chain configuration files. | `<project-root>/configs/L1` |
+| `L2_CONFIG_DIR` | Directory containing L2 chain configuration files | `<project-root>/configs/L2` |
 | `MOCK_MODE` | Whether to use mock mode | `false` |
 | `FAST_FINALITY_MODE` | Whether to use fast finality mode | `false` |
 | `RANGE_PROOF_STRATEGY` | Proof fulfillment strategy for range proofs. Set to `hosted` to use the hosted proof strategy. | `reserved` |
--- a/book/repository-structure.md
+++ b/book/repository-structure.md
@@ -2,7 +2,7 @@
 
 The OP Succinct repository structure is as follows:
 
-```
+```text
 op-succinct
 ├── contracts      # Smart contracts for OP Succinct
 │   ├── validity   # Validity contracts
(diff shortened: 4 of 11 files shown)


fault-proof/tests/sync.rs
```diff
@@ -330,4 +337,243 @@ mod sync {
...
+        for (index, (_, is_valid)) in game_sequence.iter().enumerate() {
+            let fetch_result = proposer.fetch_game(U256::from(index)).await?;
+            if *is_valid {
+                assert!(
+                    matches!(fetch_result, GameFetchResult::AlreadyExists),
+                    "Valid game at index {index} should be cached"
+                );
+            } else {
+                assert!(
+                    matches!(fetch_result, GameFetchResult::UnsupportedType { .. }),
```
Farhad-Shabani, line 413: Small nuance: the fetch result ideally should’ve returned `InvalidGame` for these legacy games, since mock games created on `TEST_GAME_TYPE` as respected type. The proposer still behaves correctly, but technically it should have surfaced the invalid game error.
Minor, up to you whether to fix now or leave for a later ticket.
But if wanted to fix, you should ensure only remove the legacy game itself afterward, not its subtree.

fault-proof/tests/sync.rs
```diff
@@ -330,4 +337,243 @@ mod sync {
...
+    }
+
+    /// Verifies that sync_state correctly filters multiple legacy games scattered in history.
+    ///
+    /// This test creates a mixed sequence of legacy and valid games, then verifies that:
+    /// - sync_state only caches valid games
+    /// - canonical head is the latest valid game
+    /// - fetch_game returns UnsupportedType for all legacy games
+    #[tokio::test]
+    async fn test_sync_state_filters_multiple_legacy_games() -> Result<()> {
```
Farhad-Shabani: BTW, this case basically shouldn’t ever happen in practice, right? I’d still like to keep it in, but maybe we should add a comment so devs don’t think it’s a normal scenario.

fault-proof/tests/sync.rs
```diff
@@ -330,4 +337,243 @@ mod sync {
...
+    async fn test_invalid_game_type_handling() -> Result<()> {
+        let (env, proposer, init_bond) = setup().await?;
+
+        // Setup legacy game type infrastructure
+        let legacy_impl = env.deploy_mock_permissioned_game().await?;
+
+        let set_init_call = DisputeGameFactory::setInitBondCall {
+            _gameType: MOCK_PERMISSIONED_GAME_TYPE,
+            _initBond: init_bond,
+        };
```
Farhad-Shabani: Now that these calls are happening more often across different tests, maybe we should turn them into methods e.g. on `env` as well so not repeating ourselves.

fault-proof/tests/sync.rs
fakedev9999: a07d5ed

fault-proof/tests/sync.rs
fakedev9999: f1aaa5e

fault-proof/tests/sync.rs
fakedev9999: @Farhad-Shabani Good catch on the error type distinction! However, the InvalidGame case is already covered in test_sync_state_filters_non_respected_game_type that I added in this PR. 

That test specifically:
- Creates a game with TEST_GAME_TYPE when MOCK_PERMISSIONED_GAME_TYPE is respected (so wasRespectedGameTypeWhenCreated = false)
- Verifies fetch_game returns InvalidGame for that scenario

This test_sync_state_filters_legacy_games test is actually testing a different scenario - filtering games of a type the proposer doesn't handle at all. So it seems like UnsupportedType might be correct for this test?

fault-proof/tests/sync.rs
```diff
@@ -330,4 +337,142 @@ mod sync {
...
+    /// This comprehensive test covers both simple (single legacy game) and complex (multiple
+    /// scattered legacy games) scenarios. It creates a mixed sequence of legacy and valid games,
+    /// then verifies that:
+    /// - sync_state only caches valid games (filters all legacy games)
+    /// - canonical head is the latest valid game
+    /// - fetch_game returns UnsupportedType for all legacy games
+    ///
+    /// **When this could happen**:
+    /// During a game type transition, a legacy proposer wasn't shut down and keeps creating
+    /// legacy-type games
```
Farhad-Shabani, line 350: Very thorough!

fault-proof/tests/sync.rs
Farhad-Shabani: I see the differences. That was a minor distinction. We’re fully covered with tests here.

PR #697, opened 2025-11-16
ci: add `just update-elf`

Farhad-Shabani:
Comes after: #696


fakedev9999:
@Farhad-Shabani Can you update verify-binaries.md to use the new just command you introduced?

PR #698, opened 2025-11-18
test(fault-proof): add bond claim eviction test

fakedev9999:
Adds bond-claim eviction coverage to fault-proof integrations tests.


fault-proof/tests/sync.rs
```diff
@@ -473,4 +473,127 @@ mod sync {
...
+        // Step 2: Set anchor if specified in test case
+        if let Some(idx) = anchor_idx {
+            env.warp_time(MAX_CHALLENGE_DURATION + 1).await?;
+            env.resolve_game(game_addresses[idx]).await?;
+            tracing::info!("✓ Resolved game {idx} as DEFENDER_WINS");
+
+            env.warp_time(DISPUTE_GAME_FINALITY_DELAY_SECONDS + 1).await?;
+            env.set_anchor_state(game_addresses[idx]).await?;
+            tracing::info!("✓ Set game {idx} as anchor");
+        }
```
Farhad-Shabani: Do we need this step? (since `anchor_idx` is always set to `None`)

fault-proof/tests/sync.rs
```diff
@@ -473,4 +473,127 @@ mod sync {
...
+    /// - Have zero credit (bonds successfully claimed)
+    /// - Are NOT protected (anchor game or canonical head)
+    ///
+    /// Protection rules tested:
+    /// - Anchor games are always retained after bond claims
+    /// - Canonical head games are always retained after bond claims
+    /// - Games that are both anchor and canonical head are retained
+    /// - Non-protected games with zero credit are evicted
+    ///
+    /// This test uses a simple 5-game linear chain (M → 0 → 1 → 2 → 3 → 4) and exercises the
```
Farhad-Shabani, line 490: Another eviction case I’m thinking about is when we have multiple branches.
Say on branch `A`, game `i` is already finalized and claimed, but on branch `B`, game `j` is still pending (for example, because it’s being challenged). And importantly, `i > j`.
In that situation, eviction for `i` would happen before `j`, so we need to make sure `j` is still retained and not evicted prematurely.

fault-proof/tests/sync.rs
fakedev9999: Oops, no we don't. I meant to drop that step when simplifying the tests.

fakedev9999: 495b9d0

fault-proof/tests/sync.rs
fakedev9999: Just added in [7d6f317](https://github.com/succinctlabs/op-succinct/pull/698/commits/7d6f317508904808883a0da96ca6d6cdd7ba59c7)

PR #702, opened 2025-11-19
test(fault-proof): add comprehensive game state marking tests

fakedev9999:
## Summary
Adds comprehensive test coverage for the proposer's game state marking logic, specifically testing the `should_attempt_to_resolve` and `should_attempt_to_claim_bond` flags which control when the proposer
should take action on games.

### New Tests

1. **`test_in_progress_games_resolution_marking`**
    - Tests `should_attempt_to_resolve` flag logic for IN_PROGRESS games
    - Covers 5 cases:
      - IN_PROGRESS + deadline not passed → false
      - DEFENDER_WINS (resolved) → false
      - IN_PROGRESS + parent resolved + deadline not passed → false
      - IN_PROGRESS + parent resolved + deadline passed + own game → true
      - IN_PROGRESS + parent not resolved + deadline passed → false

2. **`test_bond_claim_marking`**
    - Tests `should_attempt_to_claim_bond` flag logic for finalized games
    - Covers 3 cases:
      - DEFENDER_WINS + not finalized → false
      - DEFENDER_WINS + finalized + credit > 0 → true
      - DEFENDER_WINS + finalized + credit = 0 → false

### Test Infrastructure
- Added `get_game()` method (behind `e2e` feature flag) to expose internal game state for testing

### DIFF
--- a/fault-proof/tests/sync.rs
+++ b/fault-proof/tests/sync.rs
@@ -506,7 +506,6 @@ mod sync {
     /// on-chain anchor game, so it is the only game retained after eviction.
     #[tokio::test]
     async fn test_bond_claim_cache_eviction() -> Result<()> {
-        let anchor_idx: Option<usize> = None;
         let expected_retained: Vec<usize> = vec![4];
         let expected_evicted: Vec<usize> = vec![0, 1, 2, 3];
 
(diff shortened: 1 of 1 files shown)


fault-proof/tests/sync.rs
```diff
@@ -801,4 +804,215 @@ mod sync {
...
+        let game_0 = proposer.get_game(U256::from(0)).await.unwrap();
+        assert!(
+            game_0.should_attempt_to_resolve,
+            "Game 0: should_attempt_to_resolve should be true since deadline has passed"
+        );
+
+        env.resolve_game(game_addresses[0]).await?;
+        proposer.sync_state().await?;
+
+        let game_0 = proposer.get_game(U256::from(0)).await.unwrap();
```
Farhad-Shabani:
```suggestion
        let game_0 = proposer.get_game(U256::from(0)).await?;
```

Could you propogate with `?` in other places too?

fault-proof/tests/sync.rs
```diff
@@ -801,4 +804,215 @@ mod sync {
...
         Ok(())
     }
+
+    /// Tests the `should_attempt_to_resolve` flag logic for IN_PROGRESS games.
+    ///
+    /// Verifies that the flag is correctly set based on:
+    /// - Game status (IN_PROGRESS vs resolved)
+    /// - Deadline status (passed vs not passed)
+    /// - Parent resolution status (resolved vs IN_PROGRESS)
+    /// - Ownership (own game vs not own game)
```
Farhad-Shabani: This doesn’t actually cover (reject) the case where a game isn’t owned, right? That’s probably more of a permissionless-mode scenario, so it’s not a priority to add if it’s not included.

fault-proof/tests/sync.rs
```diff
@@ -801,4 +804,215 @@ mod sync {
...
+    #[tokio::test]
+    async fn test_in_progress_games_resolution_marking() -> Result<()> {
```
Farhad-Shabani, line 937: For such time-sensitive state transitions, one key thing to verify, beyond checking the general progression, is that each transition flips exactly at its boundary condition. Like It would be good to add failure/edge cases where we warp time e.g. to `MAX_CHALLENGE_DURATION` and ensure `should_attempt_to_resolve` is still false, and then confirm that it becomes true at `MAX_CHALLENGE_DURATION + 1`.

I think this category of tests would add good coverage. I’ll leave it to you whether to include them in this PR or follow up later, and whether it makes sense to convert these into `rstest` style cases to cleanly capture the boundary checks.

fault-proof/tests/sync.rs
fakedev9999: Created GRO2-1001 and removed comment in [7bf192d](https://github.com/succinctlabs/op-succinct/pull/702/commits/7bf192dc99159cd82ed95caceffc150033487260)

fault-proof/tests/sync.rs
fakedev9999: [78b39e7](https://github.com/succinctlabs/op-succinct/pull/702/commits/78b39e74ca10e2a5b227b5768303e010ea3d696d)

fault-proof/tests/sync.rs
fakedev9999: I’d prefer to defer the exact-boundary cases for now. Because the chain timestamp keeps advancing while the test is running, which I expect testing the exact boundaries of “exactly after MAX_CHALLENGE_DURATION” kind of cases very brittle.

To test the exact boundary cleanly we’d need to rework the harness to control block timestamps (e.g., stop interval mining or add an `evm_setNextBlockTimestamp` helper). Happy to tackle that follow-up once we make the time control deterministic.

fakedev9999: Opened GRO2-1002.

Farhad-Shabani: Yeah, that makes total sense. Thanks for the ticket!

fault-proof/tests/e2e.rs
```diff
@@ -453,7 +453,9 @@ mod e2e {
                 let new_game_info = factory.gameAtIndex(new_game_index).call().await?;
 
                 // Check that the new game doesn't build on the invalid chain
-                let new_game = env.fault_dispute_game(new_game_info.proxy_).await?;
+                let new_game = env
+                    .fault_dispute_game_with_role(new_game_info.proxy_, common::env::Role::Proposer)
```
Farhad-Shabani:
```suggestion
                    .fault_dispute_game(new_game_info.proxy_)
```

Please update this in some other places too, before we move on.

Farhad-Shabani (changes requested):

fault-proof/tests/sync.rs
```diff
@@ -513,36 +527,49 @@ mod sync {
...
+        // Step 2: Set anchor if specified in test case
+        if let Some(idx) = anchor_idx {
+            env.warp_time(MAX_CHALLENGE_DURATION + 1).await?;
+            env.resolve_game(game_addresses[idx]).await?;
+            tracing::info!("✓ Resolved game {idx} as DEFENDER_WINS");
+
+            env.warp_time(DISPUTE_GAME_FINALITY_DELAY_SECONDS + 1).await?;
+            env.set_anchor_state(game_addresses[idx]).await?;
+            tracing::info!("✓ Set game {idx} as anchor");
+        }
```
Farhad-Shabani: Ah, feels like there was a regression.

fault-proof/tests/e2e.rs
fakedev9999: 7a41544e0ffda665132a5b6e948dc6bb2871e567

fault-proof/tests/sync.rs
fakedev9999: Agh I think the base branch getting merged messed it up? efb21042

PR #703, opened 2025-11-19
chore: bump hokulea to hokulea-client/v1.1.1

fakedev9999:
Bumps hokulea to hokulea-client/v1.1.1 which increases max blob size.

See https://github.com/Layr-Labs/hokulea/releases/tag/hokulea-client%2Fv1.1.1 for more details.


Farhad-Shabani:
What’s causing the Docker image failures?

PR #704, opened 2025-11-20
feat(fault-proof): validate starting L2 block

fakedev9999:
## Changes 
- Add check that contract anchor L2 block does not exceed current finalized L2 block
- Improve error messaging for misconfigured startingL2BlockNumber
- Add e2e test rejecting future starting blocks

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -212,13 +212,8 @@ where
         // Initialize state with anchor L2 block number
         let anchor_l2_block = factory.get_anchor_l2_block_number(config.game_type).await?;
 
-        Self::validate_anchor_l2_block(
-            anchor_l2_block,
-            &config,
-            host.as_ref(),
-            fetcher.as_ref(),
-        )
-        .await?;
+        Self::validate_anchor_l2_block(anchor_l2_block, &config, host.as_ref(), fetcher.as_ref())
+            .await?;
 
         let initial_state =
             ProposerState { canonical_head_l2_block: Some(anchor_l2_block), ..Default::default() };


fault-proof/src/proposer.rs
```diff
@@ -211,6 +211,32 @@ where
...
+                 the current finalized L2 block ({}). This indicates:\n\
+                 1. The contract's startingL2BlockNumber is misconfigured to a future value, OR\n\
+                 2. Your L2 node is not fully synced, OR\n\
+                 3. Your L2 RPC endpoint is incorrect.\n\n\
+                 Please verify your configuration before starting the proposer.",
+                anchor_l2_block,
+                finalized_l2_block
+            ));
+        }
+
```
Farhad-Shabani, line 217: Can you make this a funciton and call it here? or even maybe a method on fetcher?

fault-proof/src/proposer.rs
```diff
@@ -211,6 +211,32 @@ where
...
+            .await?
+            .ok_or_else(|| {
+                anyhow::anyhow!(
+                    "Cannot fetch finalized L2 block number from L2 RPC: {}\n\
+                     Please check that your L2 node is running and accessible.",
+                    config.l2_rpc
+                )
+            })?;
+
+        if anchor_l2_block > U256::from(finalized_l2_block) {
```
Farhad-Shabani: Should we also prevent our config binaries from generating corrupted JSON when a user passes a `STARTING_BLOCK_NUMBER` that’s higher than the (finalized_l2_block_number -num_blocks_for_finality) ? Or at least emit a warning?

fault-proof/tests/common/env.rs
```diff
@@ -127,6 +127,60 @@ impl TestEnvironment {
         Ok(Self { game_type, private_keys, rpc_config, fetcher, anvil, deployed })
     }
 
+    /// Setup test environment with starting L2 block offset from actual finalized block
+    ///
+    /// # Arguments
+    /// * `offset` - Positive offset creates future block (for testing misconfiguration) Negative
+    ///   offset creates past block (for testing valid scenarios) Zero offset is equivalent to
+    ///   normal setup()
```
Farhad-Shabani: From what I see, in normal setup we actually set the `starting_block_number` to [`finalized_l2_block_number.saturating_sub(num_blocks_for_finality)`](https://github.com/succinctlabs/op-succinct/blob/8e3203c08ae2bbf8870fcf9445477834760dd65b/scripts/utils/bin/fetch_l2oo_config.rs#L101)

fault-proof/tests/e2e.rs
```diff
@@ -941,4 +941,27 @@ mod e2e {
 
         Ok(())
     }
+
+    // Tests that the proposer fails fast when the contract's starting L2 block number is
+    // misconfigured to a future value (e.g., 1M blocks ahead of actual finalized L2 block).
+    // This prevents the proposer from running indefinitely without creating games.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_proposer_rejects_future_starting_block() -> Result<()> {
+        let env = TestEnvironment::setup_with_starting_block_offset(1_000_000).await?;
```
Farhad-Shabani: Let’s pick an offset value that determines exactly (from) where the rejection should happen, in this case, probably 1.

fault-proof/src/proposer.rs
fakedev9999: [2dc03f7](https://github.com/succinctlabs/op-succinct/pull/704/commits/2dc03f7baf5111f341c91fdf4b93d295d462f077)

fault-proof/src/proposer.rs
fakedev9999:
```rust
let optimism_output_data: Value = data_fetcher
    .fetch_rpc_data_with_mode(
        RPCMode::L2Node,
        "optimism_outputAtBlock",
        vec![starting_block_number_hex.into()],
    )
    .await?;
```
I expect fetching output root for the config file to return error if it's such case. Do you think we should handle this proactively?

fault-proof/tests/common/env.rs
fakedev9999: The `STARTING_BLOCK_NUMBER` is always configurable as long as it's a finalized block.
For testing harness, I think current `setup_with_starting_block_offset` suits well. WDYT?

Fwiw, the default subtraction for `num_blocks_for_finality` is to ensure the user starts from a block that has already waited the full window and finalized so the starting claim can’t be challenged.

fault-proof/tests/e2e.rs
fakedev9999: [e266dd6](https://github.com/succinctlabs/op-succinct/pull/704/commits/e266dd67364a62f146c9ce4cbe505ce92c787e6c)

fault-proof/src/proposer.rs
Farhad-Shabani: I see. If it errors out in that case, don’t need then.

fault-proof/tests/common/env.rs
Farhad-Shabani: > I think current setup_with_starting_block_offset suits well. WDYT?

Yeah, on second look, that should be enough for testing. Ty!

PR #534, opened 2025-06-18
feat: fault proof deployment qol

yuwen01:
- Rename fetch_rollup_config.rs to fetch_l2oo_config.rs for clarity
- Add fetch_fault_dispute_game_config.rs script that automatically:
  - Fetches latest finalized L2 block when STARTING_L2_BLOCK_NUMBER not set
  - Calculates rollup config hash and verification keys
  - Generates opsuccinctfdgconfig.json with all deployment parameters
- Update deploy-fdg-contracts justfile command to be single-step:
  - Automatically runs config generation before deployment
  - No manual STARTING_L2_BLOCK_NUMBER configuration required
- Modify DeployOPSuccinctFDG.s.sol to read from JSON config file first, with fallback to environment variables
- Enable Solidity optimizer with via-ir to handle stack depth issues

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #705, opened 2025-11-20
fix: fail to create config when finalized L2 block is below the finality window

Farhad-Shabani:


scripts/utils/bin/fetch_fault_dispute_game_config.rs
```diff
@@ -150,10 +150,24 @@ async fn update_fdg_config() -> Result<()> {
                     None => search_start,
                 };
 
+            if finalized_l2_block_number <= num_blocks_for_finality {
```
fakedev9999, line 156:
```suggestion
            if finalized_l2_block_number < num_blocks_for_finality {
```

scripts/utils/bin/fetch_l2oo_config.rs
```diff
@@ -98,10 +98,24 @@ async fn update_l2oo_config() -> Result<()> {
                     None => search_start,
                 };
 
+            if finalized_l2_block_number <= num_blocks_for_finality {
```
fakedev9999, line 104:
```suggestion
            if finalized_l2_block_number < num_blocks_for_finality {
```

fakedev9999 (changes requested):
I think equality case is valid.

scripts/utils/bin/fetch_fault_dispute_game_config.rs
Farhad-Shabani: So it’s okay to treat the genesis block as a finalized L2 block?

fakedev9999: Yes, that's what I think.

I think not being able to prove from genesis is a different problem, more of a problem on the kona side.

Is this blocking from having sysgo powered e2e tests?

Farhad-Shabani: I agree with you on the proving side. That’s something Kona should have handled.

But from the finality standpoint, given the logic for how we determine the L2 finalized block, i.e., something is only finalized in the OP Stack’s L2-finality sense once it’s passed `DISPUTE_GAME_FINALITY_DELAY_SECONDS`, then I don’t think the genesis block should be treated as finalized at all.

fakedev9999: `BlockId::finalized` is only L1-final; we then subtract num_blocks_for_finality ourselves.

When equality hits, it means the chain is exactly one full finality window old, so block 0 has already waited the required delay. Starting at 0 in that moment is consistent with the OP finality definition.

Btw I think we consider just reverting changes in #534 using DISPUTE_GAME_FINALITY_DELAY_SECONDS to choose STARTING_L2_BLOCK_NUMBER.
Fwiw, the point of subtracting DISPUTE_GAME_FINALITY_DELAY_SECONDS is to ensure whatever block you choose has already waited the full window so the starting claim can’t be challenged. But given how Celo is planning transition, we could consider reverting the changes.

Farhad-Shabani: > When equality hits, it means the chain is exactly one full finality window old, so block 0 has already waited the required delay.

Got it. so `DISPUTE_GAME_FINALITY_DELAY_SECONDS` is meant to be an inclusive check, meaning the starting boundary counts as an accepted block. And from your context, it sounds like that's something we agreed on than being imposed by OP's specification.

Either way, we’re pretty limited here. sysgo e2e tests fails at genesis and based on what I learned from Optimism folks, generally "op-stack chains genesis state is provided as part of the RollupConfig, which is NOT part of the chain state (but rather embedded as a program's pre-state)" therefore that's why Kona does not return anything for that on fetches as well.

fakedev9999: I see. Then doesn't it make sense to add special handling and return early or emit error for genesis block?

Farhad-Shabani: Yep, the current code in this PR is already doing that. It’s set up to bail with an error including if the starting block is at genesis: https://github.com/succinctlabs/op-succinct/pull/705/files#diff-25fbdf513cf12bbb817cbfe7f30703fa3c17a0d268b426c95f6b11ff9521c979R154-R158

Farhad-Shabani: Let me add a note for that.

fakedev9999 (approved):
lgtm

scripts/utils/bin/fetch_fault_dispute_game_config.rs
Farhad-Shabani: Added. Hope works now.

PR #708, opened 2025-11-24
test: sysgo-powered e2e testing infra for fault-proof proposer

Farhad-Shabani:
Built on top of the sysgo integration introduced in succinctlabs/optimism: https://github.com/succinctlabs/optimism/pull/316

PR #701, opened 2025-11-19
test: sysgo-powered e2e testing infra for validity proposer

Farhad-Shabani:
This implementation relies on the `sysgo` integration introduced in `succinctlabs/optimism`: https://github.com/succinctlabs/optimism/pull/315

PR #711, opened 2025-11-25
chore: remove Kurtosis

Farhad-Shabani:
This is removed because the OP Labs team is no longer using or contributing to the Optimism package for Kurtosis. They’ve shifted their testing over to their sysgo-powered op-devstack instead. We’ve also been working on our own sysgo integration.

See #701 and #708 for more context.


fakedev9999 (approved):
lgtm

PR #713, opened 2025-11-26
fix: propagate error instead of unwrap across pipeline

Farhad-Shabani:
Came across a few places where the proposer panicked during my experiment with the WIP range-splitting feature.


Farhad-Shabani:
[UPDATE]
The panic only stops the proving thread, not the whole proposer.
Therefore, this PR is more about handling those cases gracefully. But since it requires an ELF update, I’m thinking we probably shouldn’t include it in v3.4.0 (?)

utils/host/src/witness_generation/preimage_witness_collector.rs
```diff
@@ -54,7 +54,7 @@ impl<P> PreimageWitnessCollector<P>
 where
     P: CommsClient + FlushableCache + Send + Sync + Clone,
 {
-    pub fn save(&self, key: PreimageKey, value: &[u8]) {
-        self.preimage_witness_store.lock().unwrap().save_preimage(key, value.to_vec());
+    pub fn save(&self, key: PreimageKey, value: &[u8]) -> PreimageOracleResult<()> {
+        self.preimage_witness_store.lock().unwrap().save_preimage(key, value.to_vec())
```
fakedev9999: nit: maybe also propagate error for lock acquisition? I can imagine mutex poisoning rarely happen.

Farhad-Shabani: [17fca6b](https://github.com/succinctlabs/op-succinct/pull/713/commits/17fca6b4cdef74ec53bb786456a7cf432aa550aa)

PR #722, opened 2025-12-02
feat: log proposer/challenger configs on loads

Farhad-Shabani:
This is helpful when reading logs from sysgo-powered e2e tests and debugging. May be useful in other contexts too!

### DIFF
--- a/.github/workflows/e2e-sysgo-tests.yml
+++ b/.github/workflows/e2e-sysgo-tests.yml
@@ -28,7 +28,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: ["./e2e/validity/...", "./e2e/faultproof/..."]
+        target: 
+          - "./e2e/validity/bootstrap/..."
+          - "./e2e/validity/proving/..."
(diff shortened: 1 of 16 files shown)


validity/src/config.rs
```diff
@@ -90,3 +90,37 @@ pub struct RequesterConfig {
     /// How long to wait before cancelling a proof request that hasn't been assigned.
     pub auction_timeout: u64,
 }
+
+impl RequesterConfig {
+    /// Log the configuration using structured tracing fields.
+    pub fn log(&self) {
+        tracing::info!(
```
fakedev9999, line 108: Good catch! https://github.com/succinctlabs/op-succinct/commit/0c08f9b6fbac830c5472f9eae7a2cc39058ea01d

validity/src/config.rs
fakedev9999, line 28: WDYT of logging this as well?

Farhad-Shabani: https://github.com/succinctlabs/op-succinct/commit/73fe8038e5ac7968cef1e076b36431e69fb228d3

PR #724, opened 2025-12-03
fix(l2oo): validate submission interval is non-zero in `updateSubmissionInterval`

Farhad-Shabani:
## Description
  
  `initialize()` validates `submissionInterval > 0`, but `updateSubmissionInterval()` did not. Setting it to zero would break `nextBlockNumber()` and allow repeated proposals at the same block number in validity mode.


fakedev9999 (approved):
Great catch on the missing validation! LGTM!

Just one comment for best practices within our team for future development:
You might want to consider using Custom Errors (e.g., error InvalidSubmissionInterval();) instead of require with a string message.
While the runtime gas difference is negligible here, using Custom Errors is generally preferred because:
1. Deployment Cost: It reduces the contract bytecode size by removing long string data.
2. Client Integration: It allows off-chain clients (like the Proposer/Challenger) to handle errors programmatically via selectors (e.g. `if (error.selector == "0x12345678")`), rather than parsing unstable string messages.

Since the existing contract already uses require strings extensively, maintaining consistency by keeping it as-is is perfectly fine.

PR #726, opened 2025-12-05
test: extend validity proposer "happy-path" e2e tests

Farhad-Shabani:
## Description

  Adds comprehensive e2e tests for the validity proposer using sysgo infra. Includes:
  
  - Add tests for single/multiple submissions and different proof intervals
  - Configurable `ValidityConfig` with expected output block calculation
  - Create `ValiditySystem` wrapper with database access for proof verification
  - Verify range proofs are contiguous and cover expected block ranges
  - Add `L2OOClient` utilities (`OutputProposal`, `WaitForLatestBlockNumber`)
  - Enable parallel execution for validity e2e tests (from `SerialT` to `ParallelT`)

Connected PR on Optimism fork: https://github.com/succinctlabs/optimism/pull/319

### DIFF
--- a/contracts/script/fp/DeployOPSuccinctFDG.s.sol
+++ b/contracts/script/fp/DeployOPSuccinctFDG.s.sol
@@ -165,10 +165,7 @@ contract DeployOPSuccinctFDG is Script, Utils {
         return registry;
     }
 
-    function deployOrGetOptimismPortal2(FDGConfig memory config, GameType gameType)
-        internal
-        returns (address payable)
-    {
+    function deployOrGetOptimismPortal2(FDGConfig memory config, GameType gameType) internal returns (address payable) {
         address payable portalAddress;
         if (config.optimismPortal2Address != address(0)) {
             portalAddress = payable(config.optimismPortal2Address);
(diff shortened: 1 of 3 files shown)


Farhad-Shabani:
I’ve got a few other cases in my list that still need test coverage, but I’m planning to handle those in a separate PR. I want to keep the scope of this one from growing any bigger.

tests/e2e/validity/proving/proving_test.go
fakedev9999, line 1: I noticed that proving/proving_test.go doesn't have a `TestMain` function, while `bootstrap/bootstrap_test.go ` does.

The proving tests create their own isolated system per test via `NewValiditySystem()`, so I assume this is intentional to allow per-test configuration flexibility.

Could you confirm this is the expected behavior?
If so, it might be worth adding a brief comment explaining why proving intentionally omits `TestMain` (e.g., each test creates its own configured system rather than sharing a pre-configured setup).

fakedev9999 (approved):
LGTM Good coverage!

tests/e2e/validity/proving/proving_test.go
Farhad-Shabani: Good catch! Yes, that's exactly right.

When I initially set up the test infra, I meant to go with simpler TestMain approach everywhere. However, I found this has limitations when tests need different configurations. Hope it now works: [766e56a](https://github.com/succinctlabs/op-succinct/pull/726/commits/766e56a78716e33ab0ec21e1109c571b35b5c866)

PR #729, opened 2025-12-06
fix(validity): correct gap detection in agg proof validation

Farhad-Shabani:
The gap check (!=) was evaluated before the overlap check (>), making the overlap condition unreachable dead code!

### DIFF
--- a/validity/src/proposer.rs
+++ b/validity/src/proposer.rs
@@ -912,8 +912,8 @@ where
     }
 
     /// Validates an aggregation proof request by checking that:
-    /// 1. There are no duplicate/overlapping range proofs
-    /// 2. There are no gaps between consecutive range proofs
+    /// 1. There are no gaps between consecutive range proofs
+    /// 2. There are no duplicate/overlapping range proofs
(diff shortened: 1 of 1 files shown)


validity/src/proposer.rs
```diff
@@ -988,10 +988,10 @@ where
                 return false;
             }
 
-            // Check for overlap (duplicate blocks)
-            if prev_proof.end_block > curr_proof.start_block {
+            // Check for gap
+            if prev_proof.end_block != curr_proof.start_block {
```
fakedev9999:
```suggestion
            if prev_proof.end_block < curr_proof.start_block {
```

I think using < might have just prevented this dead code issue. end == start implies valid contiguous proof, so maybe reordering was unnecessary.

Farhad-Shabani: Ah, true. went under the impression that we only needed to check for inequality here. [59ee6aa](https://github.com/succinctlabs/op-succinct/pull/729/commits/59ee6aa822c3b0dce7c0ed984fecd6cf98545e7d)

PR #730, opened 2025-12-09
test(eigenda): host tests + fault-proof CI coverage

Farhad-Shabani:
## Summary
  
  - Add EigenDA host utility tests for blob fetching and witness generation
  - Run fault-proof integration tests with EigenDA feature enabled in CI (in addition to Ethereum)

### DIFF
--- a/fault-proof/tests/common/constants.rs
+++ b/fault-proof/tests/common/constants.rs
@@ -40,7 +40,7 @@ pub const RANGE_VKEY_COMMITMENT: B256 = B256::ZERO; // Mock value for testing
 
 // Test configuration for L2 block offset
 // This offset is subtracted from finalized L2 block to get the starting anchor block
-pub const L2_BLOCK_OFFSET_FROM_FINALIZED: u64 = 100;
+pub const L2_BLOCK_OFFSET_FROM_FINALIZED: u64 = 200;
 
 // Test configuration for mock permissioned games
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -64,20 +72,87 @@ jobs:
...
+          bindings
+
+    - name: Install just
+      uses: extractions/setup-just@v3
+
+    - name: Setup EigenDA SRS symlinks
+      if: matrix.da == 'eigenda'
+      run: |
+        # Create symlink so the SRS file is accessible from package directories.
+        ln -sf ../../../resources utils/eigenda/host/resources
```
fakedev9999: nit: this seems quite fragile. What do you think of adding a check like [ -d resources ] || exit 1 or using $GITHUB_WORKSPACE for absolute paths

fakedev9999 (approved):
LGTM, leaving a comment for current coverage gap that I see:

Valid canoe proof deserialization -- Only invalid/malformed proofs are tested but we never test the success path with a valid proof.
We might need to use a fixture here.

.github/workflows/integration-tests.yml
Farhad-Shabani: [8aa1b44](https://github.com/succinctlabs/op-succinct/pull/730/commits/8aa1b4496d1124c7571fbc097cf1266442fe024c)

PR #733, opened 2025-12-11
fix(validity): rm redundant sort and add db client tests

Farhad-Shabani:
## Summary

- Remove redundant `sort_by_key` call in proposer since `fetch_completed_ranges` already returns results ordered by `start_block ASC` from the database query
- Add comprehensive test suite for `DriverDBClient` using embedded PostgreSQL.

### DIFF
--- a/.github/workflows/e2e-sysgo-tests.yml
+++ b/.github/workflows/e2e-sysgo-tests.yml
@@ -31,8 +31,10 @@ jobs:
         target: 
           - "./e2e/validity/bootstrap/..."
           - "./e2e/validity/proving/..."
+          - "./e2e/validity/recovery/..."
           - "./e2e/faultproof/bootstrap/..."
           - "./e2e/faultproof/fastfinality/..."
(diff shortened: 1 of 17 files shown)


fakedev9999 (approved):
LGTM, but here are some coverage gap that I see:
1. Update Operations (Status Transitions) -- The entire proof lifecycle depends on correct status updates. A bug in update_request_status could:
  - Leave requests stuck in WitnessGeneration forever
  - Allow duplicate proof requests (if status doesn't actually change)
  - Break the fetch_request_count logic that gates concurrent requests

2. Boundary Conditions (start_block == end_block) -- In the worst case of range splitting due to failure, should the insert reject if start_block = 100 and end_block = 100?

3. fetch_request_count (Concurrency Gating) -- This is the rate limiter for the entire proof system. If fetch_request_count returns wrong numbers:
  - Too low: System overloads prover with too many concurrent requests
  - Too high: System idles when it could be proving

Maybe test like test_fetch_request_count_by_status could help.

PR #734, opened 2025-12-11
test(fault-proof): clarify range splitting behavior once interval drifts

Farhad-Shabani:
## Summary
  
  - Improved docstring for `RangeSplitCount::split()` explaining how interval drift is handled
  - Added test case for drift scenario (1802 blocks instead of expected 1800)

### DIFF
--- a/fault-proof/src/config.rs
+++ b/fault-proof/src/config.rs
@@ -299,8 +299,8 @@ impl RangeSplitCount {
     /// # Behavior
     /// - Errors if `start > end` or the range is empty.
     /// - Caps the number of produced segments to the number of blocks in the range.
-    /// - Uses ceil division for even segments; may yield fewer than requested
-    ///   (e.g., 9 blocks ÷ 4 → step=3 → 3 segments: (0,3], (3,6], (6,9]).
+    /// - Uses ceil division for even segments; may yield fewer than requested (e.g., 9 blocks ÷ 4 →
+    ///   step=3 → 3 segments: (0,3], (3,6], (6,9]).
     /// - Returns ranges that exactly cover `(start, end]` with no gaps or overlaps.
     ///
     /// NOTE: At runtime, the actual interval may slightly differ from `proposal_interval_in_blocks`


fault-proof/src/config.rs
```diff
@@ -295,11 +295,14 @@ impl RangeSplitCount {
     /// - Errors if `start > end` or the range is empty.
     /// - Caps the number of produced segments to the number of blocks in the range.
     /// - Uses ceil division to keep segments as even as possible; the final segment takes any
-    ///   remainder.
-    /// - Always returns ranges that exactly cover `[start, end)` with no gaps or overlaps.
+    ///   remainder. This may yield fewer segments than requested when step sizes exhaust the range
+    ///   early (e.g., 9 blocks ÷ 4 → step=3 → 3 segments: (0,3], (3,6], (6,9]).
```
fakedev9999:
```suggestion
    ///   early (e.g., 9 blocks ÷ 4 → step=3 → 3 segments: [0,3), [3,6), [6,9)).
```
Aren't the ranges are start-inclusive, end-exclusive?

fault-proof/src/config.rs
```diff
@@ -295,11 +295,14 @@ impl RangeSplitCount {
...
+    /// - Always returns ranges that exactly cover `(start, end]` with no gaps or overlaps.
     ///
-    /// NOTE: Ceiling division may yield fewer segments than requested when step sizes exhaust the
-    /// range early. Example: 9 blocks ÷ 4 → step=3 → 3 segments: [0,3), [3,6), [6,9).
+    /// NOTE: At runtime, the actual interval may slightly differ from `proposal_interval_in_blocks`
+    /// because if a game already exists at the target L2 block, the proposer increments the block
+    /// number until it finds an unused slot (e.g., 1802 blocks instead of 1800). since we divide
+    /// the actual total by the split count, so drift just slightly adjusts range sizes.
```
fakedev9999:
```suggestion
    /// number until it finds an unused slot (e.g., 1802 blocks instead of 1800). Since we divide
    /// the actual total by the split count, drift just slightly adjusts range sizes.
```

fault-proof/src/config.rs
Farhad-Shabani: The notation is intentional. I've updated the docstring to clarify this better. [8547a85](https://github.com/succinctlabs/op-succinct/pull/734/commits/8547a85b9e1b92da952a9b42a13f4e2b7fbf2bda)
In our proving semantics, `start` is the agreed block (an already-proven checkpoint we start from) and `end` is the claimed block (what the proof verifies).

Issue #735, opened 2025-12-12
[Bug] Missing minimum interval validation in game creation allows game orphaning


louisliu2048:
### Operating System

None

## Bug Description

When multiple proposers operate with different `PROPOSAL_INTERVAL_IN_BLOCKS` configurations, games can become orphaned due to anchor state updates, causing wasted proving costs.

## Expected Behavior

Games created by honest proposers should not be orphaned by games with smaller intervals covering overlapping block ranges.

## Actual Behavior

Since `PROPOSAL_INTERVAL_IN_BLOCKS` is only a client-side configuration and the contract only validates `l2BlockNumber > parent.l2BlockNumber`, the following scenario can occur:

1. Proposer A creates game G_A covering blocks [1000 → 2800] (interval=1800)
2. Proposer B creates game G_B covering blocks [1000 → 1100] (interval=100), same parent
3. G_B completes proving faster (smaller range = faster proof)
4. G_B resolves and becomes new anchor after finality delay
5. G_A becomes orphaned because its parent is no longer the anchor

## Impact

- Proposer A loses prover fees (~$20-100) for generating proofs that will never be used
- This creates an unfair economic disadvantage for proposers using larger intervals
- Network efficiency decreases as proving work is wasted
- Malicious Proposer Can Orphan Honest Proposers' Games

## Relevant Code

**Contract validation in `initialize()`:**
```
// Only checks upper bound, no minimum interval enforcement
if (l2BlockNumber() <= startingOutputRoot.l2BlockNumber) {
    revert UnexpectedRootClaim(rootClaim());
}
```
**Client-side config (not enforced on-chain):**
```
// fault-proof/src/config.rs
pub proposal_interval_in_blocks: u64,## Suggested Improvements
```

Option 1: Add minimum interval validation in contractidity
```
uint256 constant MIN_PROPOSAL_INTERVAL = 100; // or configurable
require(l2BlockNumber() - startingOutputRoot.l2BlockNumber >= MIN_PROPOSAL_INTERVAL);
```
Option 2: Document this behavior clearly so proposers understand the risk of using large intervals in a multi-proposer environment

## Environment

- op-succinct version: latest
- Network: Any

## Additional Context

This issue becomes more significant in permissionless mode where multiple independent proposers may operate with different configurations.

PR #736, opened 2025-12-12
fix(fault-proof): conditionally allow earlier-parent games as canonical head

Farhad-Shabani:
Closes: #735

## Summary

Refactors canonical head selection to allow non-descendant games with genesis parent (`u32::MAX`) or earlier parent index to override anchor's best descendant.

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -655,8 +655,10 @@ where
 
     /// Computes the canonical head by scanning all cached games.
     ///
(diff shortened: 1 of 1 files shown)


fault-proof/src/proposer.rs
fakedev9999: Can you update the comment?

Farhad-Shabani: [0f9d981](https://github.com/succinctlabs/op-succinct/pull/736/commits/0f9d9814ba17142b5e7dc942021e08f67cc2462e)

PR #742, opened 2025-12-17
fix(tests): higher integration tests timeout

Farhad-Shabani:
Recent additions of fault-proof integration tests and EigenDA-specific tests have increased resource contention in CI, causing intermittent timeout failures. Increasing wait timeouts from `30s` to `60s` and centralizes the value in `WAIT_TIMEOUT` for easier tuning.

### DIFF
--- a/fault-proof/tests/common/constants.rs
+++ b/fault-proof/tests/common/constants.rs
@@ -46,4 +46,4 @@ pub const L2_BLOCK_OFFSET_FROM_FINALIZED: u64 = 500;
 pub const MOCK_PERMISSIONED_GAME_TYPE: u32 = 1;
 
 // Timeout (seconds) for async test operations like game creation, resolution, etc.
-pub const WAIT_TIMEOUT: u64 = 90;
+pub const WAIT_TIMEOUT: u64 = 60;


fault-proof/tests/integration.rs
```diff
@@ -944,12 +944,12 @@ mod integration {
     }
 
     // Tests that the proposer fails fast when the contract's starting L2 block number is
-    // misconfigured to a future value (e.g., a single block ahead of actual finalized L2 block).
+    // misconfigured to a future value (10 blocks ahead of the finalized block at setup time).
     // This prevents the proposer from running indefinitely without creating games.
     #[tokio::test(flavor = "multi_thread")]
     async fn test_proposer_rejects_future_starting_block() -> Result<()> {
         let env = TestEnvironment::setup_with_starting_block_offset(
-            (L2_BLOCK_OFFSET_FROM_FINALIZED + 1) as i64,
```
Farhad-Shabani, line 954: The original offset of +1 was too small. The L2 finalized block advances every ~2 seconds, so by the time contracts are deployed and the proposer initializes, the finalized block could catch up, causing the validation to pass when it should fail.

fault-proof/tests/common/constants.rs
```diff
@@ -44,3 +44,6 @@ pub const L2_BLOCK_OFFSET_FROM_FINALIZED: u64 = 500;
 
 // Test configuration for mock permissioned games
 pub const MOCK_PERMISSIONED_GAME_TYPE: u32 = 1;
+
+// Timeout (seconds) for async test operations like game creation, resolution, etc.
+pub const WAIT_TIMEOUT: u64 = 60;
```
fakedev9999, line 49: nit: wdyt of making timeout configurable via env var? This would allow CI to use longer timeouts without code changes. Low priority since the constant approach is simpler.

Farhad-Shabani: Yeah, let’s move on for now. This change was needed because of EigenDA. Otherwise, we really shouldn’t have to adjust this very often.

PR #745, opened 2025-12-19
feat(fault-proof): custom proof polling with retries, timeouts, and failure metrics

Farhad-Shabani:
## Description

Separates proof submission from status polling in the fault-proof proposer, adding robust timeout and retry handling. 
This is a preliminary step toward caching request IDs and tracking proofs across restarts.

## Summary
  - Refactor proof generation to use `request_async()` + custom polling loop instead of `run_async()`
  - Extract proving logic into new `prover.rs` module with `ProofProvider` abstraction
  - Add granular timeout handling: proving timeout, network call timeout, auction timeout
  - Add deadline check before spawning proof tasks to skip proving if deadline already passed
  - Add Prometheus metrics for timeout tracking

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -1921,6 +1921,7 @@ where
                     "Game deadline approaching, {:.1} hours remaining",
                     hours_remaining
                 );
+                ProposerGauge::DeadlineApproaching.increment(1.0);
                 Ok(false)
             }
             DeadlineStatus::Ok => Ok(false),
--- a/fault-proof/src/prometheus.rs
+++ b/fault-proof/src/prometheus.rs
@@ -103,6 +103,11 @@ pub enum ProposerGauge {
     )]
     DeadlineExceededError,
     #[strum(
+        serialize = "op_succinct_fp_deadline_approaching",
+        message = "Total number of deadline approaching warnings"
+    )]
+    DeadlineApproaching,
+    #[strum(
         serialize = "op_succinct_fp_network_call_timeout",
         message = "Total number of network call timeouts"
     )]


book/fault_proofs/proposer.md
```diff
@@ -96,7 +96,9 @@ Depending on the one you choose, you must provide the corresponding environment
 | `USE_KMS_REQUESTER` | Whether to expect NETWORK_PRIVATE_KEY to be an AWS KMS key ARN instead of a plaintext private key. | `false` |
 | `MAX_PRICE_PER_PGU` | The maximum price per pgu for proving. | `300,000,000` |
 | `MIN_AUCTION_PERIOD` | The minimum auction period (in seconds). | `1` |
-| `TIMEOUT` | The timeout to use for proving (in seconds). | `14,400` (4 hours) |
+| `TIMEOUT` | The proving timeout (in seconds). Used as the server-side deadline for proof requests and as the client-side maximum wait time when polling for proof completion. | `14,400` (4 hours) |
+| `NETWORK_CALLS_TIMEOUT` | The timeout for individual network API calls like `get_proof_status` (in seconds). If a single call exceeds this, it will be retried. | `15` |
+| `AUCTION_TIMEOUT` | The auction timeout (in seconds). If a proof request remains in "Requested" state (no prover picked it up) beyond this duration after creation, the request is cancelled. | `60` (1 minute) |
```
kien-rise: Since you’re already editing this file, I suggest splitting this into `RANGE_AUCTION_TIMEOUT` and `AGG_AUCTION_TIMEOUT`. Operators like us don’t mind adding one more environment variable. More importantly, I don’t think it’s safe to assume that `RANGE_AUCTION_TIMEOUT` is equal to `AGG_AUCTION_TIMEOUT`.

The same concern applies to `TIMEOUT`.

Farhad-Shabani: Appreciate the suggestion, Is this something necessary for your use case?
This is more of a nice-to-have. Range proofs are typically larger and longer-running, so their timeout is the primary concern. That's why a single shared timeout covers both cases adequately.
We'd like also to keep it consistent with how validity proof timeouts are handled right now. Plus we're already managing a good number of configuration variables and prefer to keep that surface area minimal - only adding new configs when truly required.

fault-proof/src/prover.rs
```diff
@@ -0,0 +1,580 @@
...
+                    proof_id,
+                    elapsed_secs,
+                    self.config.timeout
+                );
+            }
+
+            // Get proof status - retry on transient failures.
+            let (status, proof) = match self
+                .network_call_with_timeout(
+                    self.prover.get_proof_status(proof_id),
```
fakedev9999, line 219: Shouldn't get_proof_request be enough? Why do we need this call?

fault-proof/src/proposer.rs
```diff
@@ -200,17 +203,47 @@ where
...
         );
         let (range_pk, range_vk) = network_prover.setup(get_range_elf_embedded());
-        let (agg_pk, _) = network_prover.setup(AGGREGATION_ELF);
+        let (agg_pk, agg_vk) = network_prover.setup(AGGREGATION_ELF);
+
+        let keys = ProofKeys {
+            range_pk: Arc::new(range_pk),
+            range_vk: Arc::new(range_vk),
+            agg_pk: Arc::new(agg_pk),
+            agg_vk: Arc::new(agg_vk),
```
fakedev9999: Do we need the agg vkey?

We could add a preflight check like validity proposer and keep it. But now, I'm not seeing any where in the code being used.

fault-proof/src/prover.rs
```diff
@@ -0,0 +1,580 @@
...
+    NetworkProver, SP1ProofMode, SP1ProofWithPublicValues, SP1ProvingKey, SP1Stdin,
+    SP1VerifyingKey, SP1_CIRCUIT_VERSION,
+};
+use tokio::time::sleep;
+
+use crate::{config::ProofProviderConfig, prometheus::ProposerGauge};
+use op_succinct_proof_utils::get_range_elf_embedded;
+
+/// Polling interval (in seconds) for checking proof status.
+/// Matches the SP1 SDK's internal polling interval.
```
fakedev9999: Can you add a permalink to the sp1-sdk code?

fault-proof/src/proposer.rs
```diff
@@ -200,17 +203,47 @@ where
...
+        };
 
         let l1_provider = ProviderBuilder::default().connect_http(config.l1_rpc.clone());
         let l2_provider = ProviderBuilder::default().connect_http(config.l2_rpc.clone());
         let init_bond = factory.fetch_init_bond(config.game_type).await?;
+        let game_impl_address = factory.gameImpls(config.game_type).call().await?;
+        let game_impl =
+            OPSuccinctFaultDisputeGame::new(game_impl_address, factory.provider().clone());
+        let max_challenge_duration = game_impl.maxChallengeDuration().call().await?.to::<u64>();
+        let max_prove_duration = game_impl.maxProveDuration().call().await?;
```
fakedev9999: Contract params are fetched at startup only. If the contract is upgraded and these values change, the proposer would use stale values until restart. This is probably acceptable, but worth noting. Any better ways you have in mind?

fault-proof/src/proposer.rs
```diff
@@ -1835,6 +1776,50 @@ where
...
+                    game_address = ?game_address,
+                    is_defense = is_defense,
+                    "Game deadline approaching, {:.1} hours remaining",
+                    hours_remaining
+                );
+                Ok(false)
+            }
+            DeadlineStatus::Ok => Ok(false),
+        }
+    }
```
fakedev9999, line 1929: If we are handling deadline this way, I'm not seeing the point why we are not using `gameOver()` contract calls.

fault-proof/src/prover.rs
Farhad-Shabani: `get_proof_status` returns the fulfillment status and proof, but doesn't include the `created_at` timestamp. We need `get_proof_request` to get the `created_at` field, which is required to calculate whether the auction timeout has been exceeded (i.e., whether `current_time > created_at + auction_timeout` when the request is still in `Requested` state).

This is something looks like should be improved on the SDK side. Ideally `get_proof_status` would include `created_at` so we don't need two separate API calls per poll iteration.

fault-proof/src/proposer.rs
Farhad-Shabani: Right. It's not currently used. Removed for now. We can add it back when/if we implement preflight checks like the validity proposer has.
https://github.com/succinctlabs/op-succinct/commit/ffcb7ad54e3b05f9b8f0c6b349b1e8abdb1bcf6b

fault-proof/src/prover.rs
Farhad-Shabani: https://github.com/succinctlabs/op-succinct/commit/0c7e031f6c44999ef8a6a231815b26fe65b59a7d

fault-proof/src/proposer.rs
Farhad-Shabani: Good question! `should_skip_proving` (just renamed it for clarity) serves a different purpose than `gameOver()`. It's a pre-check before we start proving, plus it warns us when the deadline is approaching.

fault-proof/src/proposer.rs
Farhad-Shabani: Good observation! The impact here is low since these values are only used for edge cases (deadline checks). Plus by their nature, `max_challenge_duration` and `max_prove_duration` are params that are really rare to change. Probably fine to leave as-is for now.
Later once we add CLI support, we can add a command to trigger refreshing these params without a full restart for such a case.

book/fault_proofs/proposer.md
kien-rise: Yeah, my suggestion is just optional. Given your concern, let's keep the original env var names.

fault-proof/src/proposer.rs
fakedev9999: My question is more like: "What's the point of warning when `DeadlineStatus::Approaching`?"

If we are not taking any measures when the deadline is approaching I think it's better to make it simpler.

Farhad-Shabani: It's meant to give operators a heads-up before it's too late (`CHALLENGER_WINS`). The threshold is half of max challenge/prove duration, figured that's a reasonable point where if a range still isn't proven, something's off.

The remedy atm is manual, like tweaking some configs (lower `PROPOSAL_INTERVAL_IN_BLOCKS` or bump `RANGE_SPLIT_COUNT`) and restart the proposer, and honestly it's enough of an edge case. My thought was It doesn't worth adding self-healing logic for. But a heads-up felt warranted. 
Added a metric ([791b9de](https://github.com/succinctlabs/op-succinct/pull/745/commits/791b9de8be8ceaded37faa583e236f4e4fd0efdd)) for easier alerting too.

PR #746, opened 2025-12-19
fix(fault-proof): check respected game type before creating games

Farhad-Shabani:
Related PR: https://github.com/succinctlabs/optimism/pull/323

The proposer now verifies its configured game type matches the portal's `respectedGameType()` before creating dispute games. This ensures the proposer waits during game type transitions rather than creating games that can't finalize withdrawals.

This is caught by the `test_game_type_transition_while_proposer_running` test failure because games were being created before the transition completed.

### DIFF
--- a/fault-proof/bin/proposer.rs
+++ b/fault-proof/bin/proposer.rs
@@ -1,8 +1,6 @@
-use std::{env, sync::Arc};
+use std::sync::Arc;
 
-use alloy_primitives::Address;
 use alloy_provider::ProviderBuilder;
-use alloy_transport_http::reqwest::Url;
 use anyhow::Result;
 use clap::Parser;
 use fault_proof::{
@@ -41,7 +39,7 @@ async fn main() -> Result<()> {
 
     let proposer_signer = SignerLock::from_env().await?;
 
-    let l1_provider = ProviderBuilder::new().connect_http(proposer_config.l1_rpc);
+    let l1_provider = ProviderBuilder::new().connect_http(proposer_config.l1_rpc.clone());
 
     let anchor_state_registry = AnchorStateRegistry::new(
         proposer_config.anchor_state_registry_address,


book/fault_proofs/docker.md
```diff
@@ -26,6 +26,7 @@ The fault proof system consists of two main components:
 # Required Configuration
 L1_RPC=                  # L1 RPC endpoint URL
 L2_RPC=                  # L2 RPC endpoint URL
+PORTAL_ADDRESS=          # Address of the OptimismPortal2 contract
```
kien-rise: I think we can use AnchorStateRegistry instead:

```
cast call \
	--rpc-url "$L1_RPC_URL" \
	"$ANCHOR_STATE_REGISTRY_PROXY" \
	"respectedGameType()(uint32)"
```

This way should be easier since AnchorStateRegistry has already been used in the repo.

Farhad-Shabani: Deriving the registry from repo - using`gameImpls(game_type)` - has a migration issue: if the proposer is configured for say game type 42 but the factory hasn't registered that implementation yet, we can't get the registry address. In this case, the proposer would need to check `respectedGameType()` before its implementation is registered.

That said, you got me thinking. Maybe I add `ANCHOR_STATE_REGISTRY_ADDRESS` variable instead. It's probably better long-term since it has `isGameFinalized`, `isGameProper`, and can provide the portal via `registry.portal()` too.

kien-rise: Yes, I agree that adding `ANCHOR_STATE_REGISTRY_ADDRESS` makes sense, and it will also simplify a lot of things.

Previously, when I read the `get_anchor_state_registry_address` function, I always wondered why we needed to make several contract calls just to compute what is effectively a constant. By introducing `ANCHOR_STATE_REGISTRY_ADDRESS`, we can eliminate this function:

```rust
/// Get the anchor state registry address.
async fn get_anchor_state_registry_address(&self, game_type: u32) -> Result<Address> {
    let game_impl_address = self.gameImpls(game_type).call().await?;
    let game_impl = OPSuccinctFaultDisputeGame::new(game_impl_address, self.provider());
    let anchor_state_registry_address = game_impl.anchorStateRegistry().call().await?;
    Ok(anchor_state_registry_address)
}
```

Farhad-Shabani: Of course. Been working on that: https://github.com/succinctlabs/op-succinct/pull/748

fault-proof/tests/common/constants.rs
```diff
@@ -40,7 +40,7 @@ pub const RANGE_VKEY_COMMITMENT: B256 = B256::ZERO; // Mock value for testing
 
 // Test configuration for L2 block offset
 // This offset is subtracted from finalized L2 block to get the starting anchor block
-pub const L2_BLOCK_OFFSET_FROM_FINALIZED: u64 = 200;
+pub const L2_BLOCK_OFFSET_FROM_FINALIZED: u64 = 500;
```
fakedev9999, line 43: Could you explain why this was increased from 200 to 500?

fault-proof/bin/proposer.rs
fakedev9999: Can't we use the values from ProposerConfig?

fakedev9999 (approved):
lgtm

fault-proof/tests/common/constants.rs
Farhad-Shabani: The EigenDA node used in CI has longer confirmation times. With an offset of 200, the computed anchor block would occasionally be higher than the finalized L2 block, causing test failures. Increased to 500 for a safer margin.

fault-proof/bin/proposer.rs
Farhad-Shabani: Makes sense. 
I actually have kept that consistent with others. Fixed all together reading from the config now: [62a8d0e](https://github.com/succinctlabs/op-succinct/pull/746/commits/62a8d0e3a0819c14652a49b3f3f6868fd8da19bb)

PR #748, opened 2025-12-23
fix(fault-proof): proper pre-migration startup + consolidate anchorStateRegistry RPC calls

Farhad-Shabani:
Comes after: #746
  
## Summary

 - Fixes proposer and challenger could crash on startup if game type is not yet registered (pre-migration)
 - Removes redundant `gameImpls()` RPC calls scattered across codebase.
 - Move startup validations into `run()` loop with retry logic - services now start and wait for game type registration.

### DIFF
--- a/scripts/utils/bin/preflight.rs
+++ b/scripts/utils/bin/preflight.rs
@@ -139,10 +139,10 @@ async fn main() -> Result<()> {
         AnchorStateRegistry::new(anchor_state_registry_address, data_fetcher.l1_provider.clone());
     info!("AnchorStateRegistry at address: {anchor_state_registry_address}");
 
-    let factor_address =
+    let factory_address =
         env::var("FACTORY_ADDRESS")?.parse::<Address>().expect("FACTORY_ADDRESS must be set");
-    let factory = DisputeGameFactory::new(factor_address, data_fetcher.l1_provider.clone());
-    info!("Factory at address: {factor_address}");
+    let factory = DisputeGameFactory::new(factory_address, data_fetcher.l1_provider.clone());
+    info!("Factory at address: {factory_address}");
 
     let game_type = env::var("GAME_TYPE")?.parse::<u32>().expect("GAME_TYPE must be set");
 


fault-proof/src/challenger.rs
```diff
@@ -37,28 +39,29 @@ impl<P> OPSuccinctChallenger<P>
 where
     P: Provider + Clone,
 {
-    /// Creates a new challenger instance for testing with provided configuration.
-    pub async fn new(
+    /// Creates a new challenger instance with provided configuration.
+    pub fn new(
```
kien-rise, line 47: How about converting this function to async? That would allow us to run `startup_validations` and set `challenger_bond` directly within this function.

If linters do not allow `async fn new`, we could consider renaming the function.

kien-rise: I believe we must set `challenger_bond` before the main logic is run.

Farhad-Shabani: > How about converting this function to async? That would allow us to run `startup_validations` and set `challenger_bond` directly within this function.

This would break pre-migration startup and limit flexibility - operators can spin up the challenger before the game type is registered on-chain. I am also prefering to keep validation separate from construction for testability.

> I believe we must set challenger_bond before the main logic is run.

Correct, that's already set before any logic run. 
Btw, I just switched to `OnceLock` to make it semantically explicit too.

fault-proof/src/proposer.rs
```diff
@@ -1010,10 +1086,12 @@ where
         output_root: FixedBytes<32>,
         extra_data: Vec<u8>,
     ) -> Result<Address> {
+        let init_bond =
+            *self.init_bond.get().expect("init_bond must be set via startup_validations");
```
fakedev9999: Wdyt of returning an error instead of panicking?

fault-proof/src/challenger.rs
```diff
@@ -413,8 +482,12 @@ where
 
     pub async fn submit_challenge_transaction(&self, game: &Game) -> Result<()> {
         let contract = OPSuccinctFaultDisputeGame::new(game.address, self.l1_provider.clone());
+        let challenger_bond = *self
+            .challenger_bond
+            .get()
+            .expect("challenger_bond must be set via startup_validations");
```
fakedev9999: Wdyt of returning an error instead of panicking?

scripts/utils/bin/preflight.rs
```diff
@@ -133,17 +132,22 @@ async fn main() -> Result<()> {
...
-    );
-    info!("Factory at address: {}", factory.address());
+    let anchor_state_registry_address = env::var("ANCHOR_STATE_REGISTRY_ADDRESS")?
+        .parse::<Address>()
+        .expect("ANCHOR_STATE_REGISTRY_ADDRESS must be set");
+    let anchor_state_registry =
+        AnchorStateRegistry::new(anchor_state_registry_address, data_fetcher.l1_provider.clone());
+    info!("AnchorStateRegistry at address: {anchor_state_registry_address}");
+
+    let factor_address =
```
fakedev9999:
```suggestion
    let factory_address =
```

Farhad-Shabani: [fab13e5](https://github.com/succinctlabs/op-succinct/pull/748/commits/fab13e5a3de32b9e5eff5a542c7e2a087d1b57d2)

fault-proof/src/challenger.rs
Farhad-Shabani: [485dd69](https://github.com/succinctlabs/op-succinct/pull/748/commits/485dd6954bafd7be0945db4f3a513dd24e25db1b)

fault-proof/src/proposer.rs
Farhad-Shabani: [485dd69](https://github.com/succinctlabs/op-succinct/pull/748/commits/485dd6954bafd7be0945db4f3a513dd24e25db1b)

Issue #750, opened 2025-12-24
Missing transaction receipt status check in fault-proof


zjg555543:
## Summary

In fault-proof mode, both proposer and challenger do not verify transaction receipt status after sending transactions. This can lead to **silent failures** where transactions are included on-chain but reverted, yet the service incorrectly assumes they succeeded.

## Real-World Example:
**Reference**: [Issue #747 - Missing `l2SequenceNumber()` method in OPSuccinctFaultDisputeGame](https://github.com/succinctlabs/op-succinct/issues/747)

**Scenario**: 
- An incorrect contract was deployed (missing the `l2SequenceNumber()` method)
- When `claimCredit()` was called, it internally invoked `AnchorStateRegistry.setAnchorState()`
- `setAnchorState()` attempted to call `game.l2SequenceNumber()`, which didn't exist
- The contract call **reverted** with `EvmError: Revert`

**What Happened**:
- ✅ The transaction was **mined successfully** (included in a block)
- ❌ The transaction **execution failed** (`receipt.status = 0`)
- ✅ The proposer/challenger **logged "success"** and continued
- ❌ The anchor state was **never updated**
- ❌ The credit was **never claimed**

### Proposer (`fault-proof/src/proposer.rs`)

1. **`agg_proof_request` (Line 975-982)** - `FaultDisputeGame.prove()`
```
let receipt = self
    .signer
    .send_transaction_request(self.config.l1_rpc.clone(), transaction_request)
    .await?;

Ok(receipt.transaction_hash)  // ⚠️ No status check2. **`create_game` (Line 1086-1089)** - `DisputeGameFactory.create()`
let receipt = self
    .signer
    .send_transaction_request(self.config.l1_rpc.clone(), transaction_request)
    .await?;
// ⚠️ Directly uses receipt.inner.logs() without checking status3. **`submit_resolution_transaction` (Line 1187-1190)** - `FaultDisputeGame.resolve()`
let receipt = self
    .signer
    .send_transaction_request(self.config.l1_rpc.clone(), transaction_request)
    .await?;
// ⚠️ Logs success without verifying transaction status4. **`submit_bond_claim_transaction` (Line 1209-1212)** - `FaultDisputeGame.claimCredit()`
let receipt = self
    .signer
    .send_transaction_request(self.config.l1_rpc.clone(), transaction_request)
    .await?;
// ⚠️ Logs "Bond claimed successfully" without verification### Challenger (`fault-proof/src/challenger.rs`)
```
[...]

PR #751, opened 2025-12-24
fix(fault-proof): add transaction receipt status checks

Farhad-Shabani:
Comes after: #745
Resolves: #750

### DIFF
--- a/fault-proof/src/lib.rs
+++ b/fault-proof/src/lib.rs
@@ -195,6 +195,9 @@ where
     Ok(parent_game_contract.status().call().await? == GameStatus::CHALLENGER_WINS)
 }
 
+/// Prefix used for transaction revert errors.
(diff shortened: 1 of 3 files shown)


fault-proof/src/lib.rs
```diff
@@ -240,3 +241,15 @@ where
 
     Ok(parent_game_contract.status().call().await? == GameStatus::CHALLENGER_WINS)
 }
+
+/// Extension trait for checking transaction error types.
+pub trait TxErrorExt {
```
fakedev9999, line 202: The current implementation seems very fragile. Wdyt of adding a custom error type e.g., `pub enum TxError`?

Farhad-Shabani: I see you point. I actually considered a custom enum but it would require changing many function signatures for what's essentially a binary check used only for log-level selection.
Added a `TX_REVERTED_PREFIX` constant. Hope feels better now.

PR #756, opened 2025-12-30
feat(fault-proof): proposer state backup for faster recovery

Farhad-Shabani:
Related PR: https://github.com/succinctlabs/optimism/pull/325

## Summary

Adds **optional** file-based backup for proposer state. On restart, the proposer restores its game cache from disk instead of re-syncing from the factory, reducing recovery time and RPC load.

## Fallback Behavior

Designed so to never disrupt normal operation:
- Load failures (missing file, parse error, version mismatch, validation failure) → proposer starts fresh
- Save failures → logged as warning, main loop unaffected


fault-proof/src/backup.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// Load and validate a backup from file.
+    ///
+    /// Returns None and logs a warning if:
+    /// - File doesn't exist or can't be read
+    /// - JSON parsing fails
+    /// - Version mismatch
+    /// - Validation fails (stale/corrupted data)
+    pub fn load(path: &Path) -> Option<Self> {
```
fakedev9999, line 89: Wdyt of failing hard on load errors? Or maybe at least add metrics so operators have better observability?

fault-proof/src/proposer.rs
```diff
@@ -318,6 +322,8 @@ where
                 continue
             }
 
+            self.backup().await;
```
fakedev9999, line 327: The backup is triggered every fetch cycle. With FETCH_INTERVAL=1, this attempts a backup every second. While the semaphore prevents concurrent backups, the filesystem I/O overhead may accumulate, especially on slower storage.

Two possible improvements (either/both):

1. Time-based throttling: Only backup every N seconds.
2. Dirty flag: Only backup when state actually changed.

Not blocking - the current implementation works fine. Just flagging for consideration if high-frequency polling becomes common.

fault-proof/src/proposer.rs
```diff
@@ -2083,6 +2119,45 @@ pub fn check_deadline_status(now: u64, deadline: u64, max_duration: u64) -> Dead
...
+        let games: HashMap<U256, Game> = backup.games.into_iter().map(|g| (g.index, g)).collect();
+
+        let anchor_game = backup.anchor_game_index.and_then(|idx| games.get(&idx).cloned());
+
+        Self {
+            cursor: backup.cursor.map(Cursor::from).unwrap_or_default(),
+            games,
+            anchor_game,
+            canonical_head_index: None,
+            canonical_head_l2_block: None,
```
fakedev9999, line 2168: Could you add a brief comment explaining that these are intentionally not persisted because they're re-computed on the next sync cycle?

fault-proof/src/backup.rs
```diff
@@ -0,0 +1,124 @@
...
+use std::{collections::HashSet, io::Write, path::Path};
+
+use alloy_primitives::U256;
+use anyhow::{bail, Context, Result};
+use serde::{Deserialize, Serialize};
+
+use crate::proposer::Game;
+
+/// Current backup format version. Increment when making breaking changes.
+pub const BACKUP_VERSION: u32 = 1;
```
fakedev9999, line 17: The manual `BACKUP_VERSION` constant works but requires developers to remember to increment it. I think this is fragile.

A more robust approach could use compile-time type hashing using [type_hash](https://crates.io/crates/type_hash) like crates.

fault-proof/src/proposer.rs
```diff
@@ -413,6 +419,17 @@ where
         self.contract_params
             .set(contract_params)
             .map_err(|_| anyhow::anyhow!("contract_params must not already be set"))?;
+
+        // Restore state from backup if available.
+        if let Some(path) = &self.config.backup_path {
```
fakedev9999, line 426: I think it'd be great to have backup path validation some where early so the error surfaces early rather than on first backup attempt.

fault-proof/src/backup.rs
```diff
@@ -0,0 +1,124 @@
...
+        let json =
+            serde_json::to_string_pretty(self).context("failed to serialize proposer backup")?;
+
+        let tmp_path = path.with_extension("tmp");
+
+        // Write and sync to ensure data is on disk before rename.
+        let mut file = std::fs::File::create(&tmp_path)
+            .context("failed to create proposer backup temp file")?;
+        file.write_all(json.as_bytes()).context("failed to write proposer backup temp file")?;
+        file.sync_all().context("failed to sync proposer backup temp file")?;
```
fakedev9999: Currently, the temp file is only cleaned up on rename failure. If `write_all` or `sync_all` fails, the temp file is left behind. Consider adding additional cleanups.

Not critical since the .tmp file won't interfere with anything, but cleaner.

fault-proof/src/backup.rs
```diff
@@ -0,0 +1,124 @@
...
+        drop(file);
+
+        // Rename with cleanup on failure.
+        if let Err(e) = std::fs::rename(&tmp_path, path) {
+            let _ = std::fs::remove_file(&tmp_path);
+            return Err(e).context("failed to rename proposer backup file");
+        }
```
fakedev9999: Consider using [tempfile::NamedTempFile::persist()](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.persist) since `persist()` handles atomic rename + cleanup on failure.

Not blocking - current implementation is correct, just more verbose.

PR #758, opened 2025-12-31
feat(fault-proof): add vkey, rollup config hash validation to prevent proving incompatible games

fakedev9999:
## Summary

This PR adds vkey and rollup config hash validation plus prove-only mode to enable safe proposer operation during hardfork transitions when ELF files and verification keys change.

## Problem

During hardfork transitions:
- ELF files and verification keys (vkeys) are updated
- Old games retain their original vkeys and rollup config hash (immutable on-chain)
- New proposers with updated ELFs cannot prove old games (vkey mismatch)
- Proofs submitted to incompatible games fail on-chain verification, wasting compute and gas

### 1. Identity Validation

Proposer computes identity parameters from ELF and chain config at startup:
- `aggregation_vkey` - from aggregation ELF
- `range_vkey_commitment` - from range ELF
- `rollup_config_hash` - from rollup config

Before proving any game, validates all three on-chain parameters match. Skips games with mismatched identity with clear warning logs.

### 2. Proposer Identity Tracking

New `ProposerIdentity` struct with version info (Cargo version + DA layer) and identity parameters. Startup logging shows version and identity fingerprints for operator visibility:

```
version=3.4.1-ethereum aggregation_vkey=0xabcd... range_vkey=0x1234... rollup_config_hash=0x5678...
```

## Dependencies

- succinctlabs/optimism#336 - Adds `UpgradeGameImpl` method to sysgo for Forge script deployment

## Test plan

- [x] E2E test validates proposer skips game creation on vkey mismatch
- [x] E2E test uses same Forge script (`UpgradeOPSuccinctFDG.s.sol`) that operators use in production

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -16,10 +16,7 @@ use alloy_provider::{Provider, ProviderBuilder};
 use alloy_sol_types::{SolEvent, SolValue};
 use anyhow::{bail, Context, Result};
 use futures::stream::{self, StreamExt, TryStreamExt};
-use op_succinct_client_utils::{
-    boot::{hash_rollup_config, BootInfoStruct},
-    types::u32_to_u8,
-};
+use op_succinct_client_utils::boot::{hash_rollup_config, BootInfoStruct};
 use op_succinct_elfs::AGGREGATION_ELF;
 use op_succinct_host_utils::{
     fetcher::OPSuccinctDataFetcher,
@@ -300,15 +297,8 @@ where
 
         // Compute vkey hashes for on-chain compatibility checks.
         // These are compared against game contract immutable values to ensure proof compatibility.
-        let aggregation_vkey = {
-            let hex_str = agg_vk.bytes32();
-            // bytes32() returns "0x..." prefixed hex string
-            B256::from_slice(
-                &hex::decode(hex_str.trim_start_matches("0x"))
-                    .context("failed to decode aggregation vkey hex")?,
-            )
-        };
-        let range_vkey_commitment = B256::from(u32_to_u8(range_vk.vk.hash_u32()));
+        let aggregation_vkey = B256::from(agg_vk.bytes32_raw());
+        let range_vkey_commitment = B256::from(range_vk.hash_bytes());
 
         // Compute rollup config hash from the fetcher's chain config.
         let rollup_config_hash = hash_rollup_config(


fakedev9999:
> Took a look at the PR. Nice work on the core pieces (vkey compatibility checks, ProposerIdentity). Before I share some thoughts, wanted to make sure I understand the intended flow:
> 
> ### Quick clarification
> The current approach:
> 
> * Operator runs old proposer normally
> * Hardfork happens → operator restarts old proposer with `PROVE_ONLY_MODE=true`
> * Operator starts new proposer (creates new games)
> * Old proposer winds down (only proves/resolves/claims existing games)
> 
> Is that right?
> 
> ### Alternative: Self-adapting
> If so, here's what I've been thinking about: What if the proposer just figured this out automatically? Since we already have the vkeys from the contract, the proposer always knows if it's compatible or not.
> 
> The idea:
> 
> 1. Before creating a game, check if factory's vkeys match ours
> 2. If they match → create games normally
> 3. If they don't match → skip game creation, but keep doing everything else (proving, resolving, claiming)
> 
> So during a hardfork:
> 
> * Old proposer: automatically stops creating games (vkey mismatch), but keeps proving its existing games
> * New proposer: starts creating games (vkey match), builds on top of whatever the old proposer left
> 
> No need for restart, flag or operator intervention. The proposers just figure it out from chain state. And if an old game gets challenged while both are running? The existing `RemoveSubtree` logic handles it. New proposer sees the reorg, its games built on top of the invalid one get removed, and it rebuilds from the last valid game. Both proposers stay in sync through chain state.
> 
> ### Owned vs Foreign games
> One enhancement though we need: Instead of caching ALL games the same way, we could classify them:
> 
> * Owned = vkeys match, we can prove these
> * Foreign = vkeys don't match, just tracking for chain continuity
> 
> Each proposer only operates on Owned games - proving, resolving, and claiming bonds. Foreign games are just tracked for DAG continuity, not acted upon. This also makes state persistence portable (#756). New proposer can load old proposer's backup, reclassify everything based on its own vkeys, and pick up right where it left off.
> 
> WDYT? Does this simplify things or am I missing something about why the explicit mode flag is needed?

Yes, your understanding is correct and your self-adapting approach is cleaner - I'm on board.

I think it's better because chain state is the source of truth.

fault-proof/src/proposer.rs
```diff
@@ -1827,21 +1969,36 @@ where
...
         deadline: Option<u64>,
     ) -> Result<()> {
+        // Skip if deadline has passed
         if let Some(deadline) = deadline {
             if self.should_skip_proving(game_address, deadline, is_defense)? {
                 return Ok(());
             }
         }
 
+        // Only prove owned games.
```
Farhad-Shabani: I think this check belongs in the `should_skip_proving` method. It decides whether proving should be skipped, so it feels like the right place for it.

fault-proof/tests/common/constants.rs
```diff
@@ -34,9 +34,9 @@ pub const MAX_PROVE_DURATION: u64 = 60 * 60 * 12; // 12 hours
 pub const FALLBACK_TIMEOUT: U256 = U256::from_limbs([1209600, 0, 0, 0]); // 2 weeks
 
 // Configuration hashes for OPSuccinctFaultDisputeGame
-pub const ROLLUP_CONFIG_HASH: B256 = B256::ZERO; // Mock value for testing
-pub const AGGREGATION_VKEY: B256 = B256::ZERO; // Mock value for testing
-pub const RANGE_VKEY_COMMITMENT: B256 = B256::ZERO; // Mock value for testing
+// Note: AGGREGATION_VKEY, RANGE_VKEY_COMMITMENT, and ROLLUP_CONFIG_HASH are now computed
+// at runtime via compute_vkeys() and hash_rollup_config() in env.rs to match the proposer's
```
Farhad-Shabani: Is this comment needed?

fault-proof/src/proposer.rs
```diff
@@ -859,6 +967,21 @@ where
         }
     }
 
+    /// Returns true if on-chain vkeys match ours (safe to create games).
+    async fn on_chain_vkeys_match(&self) -> Result<bool> {
```
Farhad-Shabani, line 991: `on_chain_vkeys_match()` checks 2 fields (agg_vkey, range_vkey) but `is_owned()` checks 3 (includes rollup_config_hash). Is that difference intentional?

tests/presets/faultproof.go
```diff
@@ -297,3 +316,217 @@ func NewDefaultFaultProofSystem(t devtest.T) *FaultProofSystem {
...
+		MaxChallengeDuration: maxChallengeDuration,
+		MaxProveDuration:     maxProveDuration,
+		ChallengerBondWei:    challengerBond,
+	}, nil
+}
+
+// DeployGameImplWithFakeVkeys deploys a new game implementation with fake vkeys.
+// This simulates a hardfork scenario where the on-chain vkeys don't match the proposer's computed vkeys.
+// The new implementation uses the same configuration as the current one, except for the vkeys.
+func (s *FaultProofSystem) DeployGameImplWithFakeVkeys(
```
Farhad-Shabani: Should we be writing deployment logic in Go at all? Operators will use Forge scripts for upgrades, so our e2e tests should use those same scripts.

Ideally `DeployGameImplWithFakeVkeys` just shells out to the existing deployment script with override params. Keeps tests aligned with what operators actually do, and we get coverage on the scripts themselves for free.

contracts/script/fp/DeployGameImplOnly.s.sol
Farhad-Shabani, line 1: I might be missing something, but couldn't we just use `UpgradeOPSuccinctFDG.s.sol` for this test? It already deploys a new impl. We'd just pass fake vkeys as env vars. 

More broadly, how are operators supposed to handle this transition?
Does `UpgradeOPSuccinctFDG.s.sol` cover it end-to-end, or does it need updating? If it doesn’t, should we add a new script instead?
Either way, we can probably reuse whichever path we pick for our tests too.

fault-proof/src/proposer.rs
```diff
@@ -77,6 +80,74 @@ pub enum TaskInfo {
...
+/// This helps operators identify which ELF version is running and enables
+/// compatibility checks during hardfork transitions.
+#[derive(Clone, Debug)]
+pub struct ProposerIdentity {
+    /// Package version from Cargo.toml (e.g., "3.4.1")
+    pub version: &'static str,
+    /// DA layer feature (e.g., "ethereum", "celestia", "eigenda")
+    pub da_layer: &'static str,
+    /// Full version string with DA layer suffix (e.g., "3.4.1-celestia")
+    pub full_version: String,
```
Farhad-Shabani: Looks like only `full_version` is actually used. `version` and `da_layer` are just intermediate values in the constructor. Could we drop them from the struct and keep them as local variables in new()? And maybe renaming `full-verison` to `version` then?

fault-proof/src/backup.rs
```diff
@@ -14,7 +14,8 @@ use serde::{Deserialize, Serialize};
 use crate::proposer::Game;
 
 /// Current backup format version. Increment when making breaking changes.
-pub const BACKUP_VERSION: u32 = 1;
+/// v2: Added aggregation_vkey, range_vkey_commitment, rollup_config_hash to Game.
+pub const BACKUP_VERSION: u32 = 2;
```
Farhad-Shabani: We can keep it at “1” since we didn’t release any new version since the backup feature was added, meaning there’s no breaking change yet.

fault-proof/src/proposer.rs
```diff
@@ -219,7 +296,26 @@ where
...
         let (range_pk, range_vk) = network_prover.setup(get_range_elf_embedded());
-        let (agg_pk, _) = network_prover.setup(AGGREGATION_ELF);
+        let (agg_pk, agg_vk) = network_prover.setup(AGGREGATION_ELF);
+
+        // Compute vkey hashes for on-chain compatibility checks.
+        // These are compared against game contract immutable values to ensure proof compatibility.
+        let aggregation_vkey = {
+            let hex_str = agg_vk.bytes32();
+            // bytes32() returns "0x..." prefixed hex string
+            B256::from_slice(&hex::decode(hex_str.trim_start_matches("0x")).unwrap())
```
Farhad-Shabani: Let's not use `unwrap`

justfile
```diff
@@ -157,6 +157,38 @@ deploy-mock-verifier env_file=".env":
     --broadcast \
     $VERIFY
 
+# Upgrade the game implementation contract (for hardfork/upgrade)
+# This script deploys a new OPSuccinctFaultDisputeGame implementation and sets it in the factory.
+# Required env vars: FACTORY_ADDRESS, GAME_TYPE, VERIFIER_ADDRESS, ANCHOR_STATE_REGISTRY, ACCESS_MANAGER,
+#                    AGGREGATION_VKEY, RANGE_VKEY_COMMITMENT, ROLLUP_CONFIG_HASH,
+#                    MAX_CHALLENGE_DURATION, MAX_PROVE_DURATION, CHALLENGER_BOND_WEI
+upgrade-game-impl env_file=".env":
```
Farhad-Shabani, line 165: 👌🏻

fault-proof/tests/common/env.rs
```diff
@@ -81,10 +85,34 @@ impl Drop for TestEnvironment {
...
 
+/// Compute vkeys from ELF programs.
+/// Returns (aggregation_vkey, range_vkey_commitment) as B256 values.
+/// Uses the same computation as `proposer.rs` and `config_common.rs`.
+pub fn compute_vkeys() -> (B256, B256) {
+    let prover = ProverClient::builder().cpu().build();
+
+    let (_, agg_vk) = prover.setup(AGGREGATION_ELF);
+    let aggregation_vkey = {
+        let hex_str = agg_vk.vk.bytes32();
```
Farhad-Shabani: Nitpicking, to stay consistent with `fault-proof/src/proposer.rs:304`:

```suggestion
        let hex_str = agg_vk.bytes32();
```

fault-proof/src/proposer.rs
```diff
@@ -219,7 +296,29 @@ where
...
+        // Compute vkey hashes for on-chain compatibility checks.
+        // These are compared against game contract immutable values to ensure proof compatibility.
+        let aggregation_vkey = {
+            let hex_str = agg_vk.bytes32();
+            // bytes32() returns "0x..." prefixed hex string
+            B256::from_slice(
+                &hex::decode(hex_str.trim_start_matches("0x"))
+                    .context("failed to decode aggregation vkey hex")?,
+            )
+        };
```
kien-rise:
```rust
let aggregation_vkey = B256::from(agg_vk.bytes32_raw())
```

fault-proof/src/proposer.rs
```diff
@@ -219,7 +296,29 @@ where
...
+        let range_vkey_commitment = B256::from(u32_to_u8(range_vk.vk.hash_u32()));
```
kien-rise:
```rust
let range_vkey_commitment = B256::from(range_vk.hash_bytes())
```

fault-proof/src/proposer.rs
```diff
@@ -1886,19 +2013,16 @@ where
...
         is_defense: bool,
         deadline: Option<u64>,
     ) -> Result<()> {
-        if let Some(deadline) = deadline {
-            if self.should_skip_proving(game_address, deadline, is_defense)? {
-                return Ok(());
-            }
+        // Skip if game is not owned or deadline has passed
+        if self.should_skip_proving(game_address, deadline, is_defense).await? {
+            return Ok(());
```
kien-rise, line 2025: The issue with returning `Ok(())` is that the caller still increments `active_defense_tasks_count += 1`, even though no task was actually spawned. As a result, a **single** unproven game with mismatched vkeys can block the proposer indefinitely (assume `self.config.max_concurrent_defense_tasks` is 1 for simplicity)

fakedev9999: Fixed in https://github.com/succinctlabs/op-succinct/pull/794. Thanks!

PR #759, opened 2026-01-01
fix(fault-proof): challenger `sync_state` update and deadline check

Farhad-Shabani:
## Description
Fixes two bugs in challenger's `sync_state` for `IN_PROGRESS` games:

1. Cached state not updated: Update action was only pushed when specific conditions were met, missing on-chain state changes like Unchallenged → Challenged transitions.

2. Challenge attempted after deadline: Missing gameOver() check for Unchallenged games caused challenger to attempt challenges that would revert on-chain.

### DIFF
--- a/fault-proof/tests/integration.rs
+++ b/fault-proof/tests/integration.rs
@@ -946,7 +946,7 @@ mod integration {
     }
 
     // Tests that the proposer fails fast when the contract's starting L2 block number is
-    // misconfigured to a future value (100 blocks ahead of the finalized block at setup time).
+    // misconfigured to a future value (100 blocks beyond of the finalized block at setup time).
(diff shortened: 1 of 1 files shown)


fault-proof/tests/integration.rs
```diff
@@ -946,12 +946,12 @@ mod integration {
     }
 
     // Tests that the proposer fails fast when the contract's starting L2 block number is
-    // misconfigured to a future value (10 blocks ahead of the finalized block at setup time).
+    // misconfigured to a future value (100 blocks ahead of the finalized block at setup time).
```
fakedev9999: nit: Unrelated to the diff but 
```suggestion
    // misconfigured to a future value (100 blocks beyond of the finalized block at setup time).
```
seems correct.

PR #770, opened 2026-01-12
feat: allow to customize RUST_LOG

kien-rise:
This PR gives users more control over logging. Previously, logs marked as “hidden” were always suppressed, with no way to override this behavior.


fakedev9999:
Superseded by https://github.com/succinctlabs/op-succinct/pull/860

PR #771, opened 2026-01-12
perf: use l2_block.header.withdrawals_root to optimize performance

kien-rise:
This improves the performance of the game creation process once the Isthmus hard fork is activated.

Generally, the game creation should avoid calling the expensive `eth_getProof`.

### DIFF
--- a/fault-proof/src/lib.rs
+++ b/fault-proof/src/lib.rs
@@ -85,12 +85,18 @@ impl L2ProviderTrait for L2Provider {
             .await?;
         let l2_state_root = l2_block.header.state_root;
         let l2_claim_hash = l2_block.header.hash;
-        let l2_storage_root = self
-            .get_l2_storage_root(
+
+        let l2_storage_root = if let Some(root) = l2_block.header.withdrawals_root {
+            // Post-Isthmus: Use the withdrawals_root from the header (fast)
+            root
+        } else {
+            // Pre-Isthmus: Fall back to eth_getProof (slow)
+            self.get_l2_storage_root(
                 address!("0x4200000000000000000000000000000000000016"),
                 BlockNumberOrTag::Number(l2_block_number.to::<u64>()),
             )
-            .await?;
+            .await?
+        };
 
         let l2_claim_encoded = L2Output {
             zero: 0,


fakedev9999 (commented):
Thanks for the optimization — skipping `eth_getProof` when the value is already in the header is a good idea.

However, checking `if let Some(root)` isn't safe here. On OP Stack L2s, `withdrawals_root` is `Some` starting from **Canyon**, but it only contains the L2ToL1MessagePasser storage root starting from **Isthmus**:

| Hardfork | `withdrawalsRoot` |
|---|---|
| Pre-Canyon | `None` |
| Canyon → pre-Isthmus | `Some(EMPTY_WITHDRAWALS_ROOT)` — MPT root of an empty withdrawals list |
| Isthmus+ | `Some(<L2ToL1MessagePasser storage root>)` |

The Isthmus exec-engine spec explicitly warns against this:

> Prior to Isthmus activation, even if a `withdrawalsRoot` is present and a MPT root is present in the header, **it should not be used**. Any implementation that calculates output root should be careful not to use the header `withdrawalsRoot`.

([source](https://specs.optimism.io/protocol/isthmus/exec-engine.html))

For any block between Canyon and Isthmus activation, this would use the empty root as `l2_storage_hash`, producing an incorrect output root. This also affects historical block lookups during dispute games, even if the chain is currently post-Isthmus.

A minimal fix would be to guard against the empty root:

```rust
const EMPTY_WITHDRAWALS_ROOT: B256 = b256!("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421");

let l2_storage_root = match l2_block.header.withdrawals_root {
    Some(root) if root != EMPTY_WITHDRAWALS_ROOT => root,
    _ => self.get_l2_storage_root(
        address!("0x4200000000000000000000000000000000000016"),
        BlockNumberOrTag::Number(l2_block_number.to::<u64>()),
    ).await?,
};
```

Alternatively, checking the block timestamp against the Isthmus activation time from the rollup config would be more precise.

kien-rise:
Your fix looks good. Thanks for pointing out the missing logic.

fakedev9999:
Superseded by https://github.com/succinctlabs/op-succinct/pull/858

PR #773, opened 2026-01-13
fix(scripts): use CLI --default-range argument instead of hardcoded constant

fakedev9999:
## Summary
- Fix bug where `--default-range` CLI argument was parsed but ignored in `multi` binary
- The hardcoded `DEFAULT_RANGE` constant was being used instead of `args.default_range`

## Test plan
- [x] `cargo build` compiles successfully
- [x] `cargo clippy` passes with no warnings
- [x] `--help` output shows `--default-range` argument with default value

### DIFF
--- a/scripts/prove/src/lib.rs
+++ b/scripts/prove/src/lib.rs
@@ -5,7 +5,6 @@ use op_succinct_host_utils::fetcher::{BlockInfo, OPSuccinctDataFetcher};
 use op_succinct_proof_utils::get_range_elf_embedded;
 use sp1_sdk::{ExecutionReport, ProverClient, SP1Stdin};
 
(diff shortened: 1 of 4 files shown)


scripts/prove/bin/multi.rs
```diff
@@ -5,7 +5,7 @@ use op_succinct_host_utils::{
     stats::ExecutionStats, witness_generation::WitnessGenerator,
 };
 use op_succinct_proof_utils::{get_range_elf_embedded, initialize_host};
-use op_succinct_prove::{execute_multi, DEFAULT_RANGE};
```
Farhad-Shabani, line 8: Maybe replace `DEFAULT_RANGE` under the `scripts/prove/tests/common/mod.rs`? since now it's only used for tests.

PR #776, opened 2026-01-14
feat(scripts): add witness caching for multi and cost_estimator

fakedev9999:
## Summary

Adds witness caching to skip the 2-hour witness generation bottleneck in `multi.rs` and `cost_estimator` scripts. Cache files are saved to `data/{chain_id}/witness-cache/{start_block}-{end_block}.bin` using rkyv serialization.

## Changes

- New witness_cache module with save/load/exists functions supporting conditional EigenDA compilation
- Three cache flags: `--cache` (recommended), `--use-cache` (load only), `--save-cache` (force regenerate)
- Integration in multi.rs and cost_estimator with per-range caching
- Comprehensive documentation on cache usage, location, and DA compatibility
- Cache files are NOT compatible across DA types (Ethereum/Celestia vs EigenDA)

## Testing

Run multi or cost_estimator with `--cache` flag to test caching behavior. Verify cache files are created at expected location.

### DIFF
--- a/.github/workflows/spell-checker.yml
+++ b/.github/workflows/spell-checker.yml
@@ -27,7 +27,8 @@ jobs:
         id: check_for_typos
         uses: reviewdog/action-misspell@v1
         with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          reporter: github-pr-review
           path: "./book"
           locale: "US"
           exclude: |
--- a/contracts/foundry.toml
+++ b/contracts/foundry.toml
@@ -19,6 +19,7 @@ remappings = [
     "src/L2/=lib/optimism/packages/contracts-bedrock/src/L2/",
     "src/dispute/=lib/optimism/packages/contracts-bedrock/src/dispute/",
     "src/cannon/=lib/optimism/packages/contracts-bedrock/src/cannon/",
+    "src/universal/=lib/optimism/packages/contracts-bedrock/src/universal/",
     "interfaces/=lib/optimism/packages/contracts-bedrock/interfaces/",
     "@lib-keccak/=lib/lib-keccak/contracts/lib",
 ]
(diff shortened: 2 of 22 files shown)


scripts/prove/bin/multi.rs
```diff
@@ -35,16 +43,51 @@ async fn main() -> Result<()> {
...
+    // Check cache first if enabled (with graceful fallback)
+    let (sp1_stdin, witness_generation_duration) = if should_try_cache {
+        match load_stdin_from_cache(l2_chain_id, l2_start_block, l2_end_block) {
+            Ok(Some(stdin)) => {
+                println!("Loaded stdin from cache");
+                (stdin, Duration::ZERO)
+            }
+            Ok(None) => generate_stdin().await?,
+            Err(e) => {
+                eprintln!("Warning: Failed to load cache: {e}, regenerating...");
```
Farhad-Shabani: What do you think about using log levels instead? The same in other spots where we’re currently using `println!`

utils/host/src/witness_cache.rs
```diff
@@ -0,0 +1,67 @@
+//! SP1Stdin caching module for saving/loading proving inputs to/from disk.
+//!
+//! This module provides functions to cache SP1Stdin (proving input), keyed by (chain_id,
+//! start_block, end_block). Caching allows skipping the time-consuming witness generation step
+//! (`host.run()`) on subsequent runs.
+//!
+//! SP1Stdin is DA-agnostic (same type regardless of witness generator), so cache files work
+//! across all DA types (Ethereum, Celestia, EigenDA).
```
Farhad-Shabani: Is this accurate statement?

My understating: `SP1Stdin` is the same type, but the contents are DA-specific. Cache files are only valid for the same DA type they were generated with.

scripts/utils/src/lib.rs
```diff
@@ -15,9 +15,16 @@ pub struct HostExecutorArgs {
...
+    /// Load witness from cache if available (skip witness generation).
     #[arg(long)]
     pub use_cache: bool,
+    /// Save generated witness to cache for future runs.
+    #[arg(long)]
+    pub save_cache: bool,
+    /// Enable caching: load from cache if available, save to cache if not.
+    /// Equivalent to --use-cache --save-cache.
+    #[arg(long)]
+    pub cache: bool,
```
Farhad-Shabani: Honestly, I think `--cache` alone is probably sufficient for most of use cases. I'd keep it simple.

b/c if a range already stored/available, it would know to use that, and if not it generates, then stores.

Farhad-Shabani (approved):
✨

PR #778, opened 2026-01-26
feat: bump kona to v1.2.7, hokulea to v1.1.4, alloy to 1.4.x

fakedev9999:
## Summary
- Upgrade kona from v1.2.2 to v1.2.7
- Upgrade hokulea from v1.1.3 to v1.1.4
- Upgrade hana to v1.5.0-mocha
- Upgrade alloy from =1.1.3 to 1.4.x for compatibility with hokulea's sp1-contract-call deps
- Replace kona-rpc with local RPC types to avoid rollup-boost dependency chain

### hokulea v1.1.4 brings:
- Extracted `canoe_proof_stdin`, `generate_canoe_proof` functions
- Bug fix for `HintType::L2PayloadWitness` proactive hints
- reth v1.9.1 → v1.9.3, SP1-CC updated
- Experimental witness endpoint enabled
- Proxy updated to v2.6.0 with V4 certificate support

### alloy version upgrade
alloy is now at 1.4.x (from =1.1.3) because hokulea v1.1.4 → sp1-contract-call → rsp requires alloy 1.4.3.

### kona-rpc removal
kona-rpc was removed because it depends on rollup-boost → reth-optimism-primitives v1.9.3, which expects alloy 1.0.x. This conflicts with alloy 1.4.x. Since we only use `OutputResponse` and `SafeHeadResponse` types from kona-rpc, these are now defined locally in `utils/host/src/rpc_types.rs`.

### reth patches
The reth patches remain in place for sp1-contract-call's reth dependencies which still need the PartialEq bug fix from [paradigmxyz/reth#21336](https://github.com/paradigmxyz/reth/pull/21336).

## Test plan
- [x] `cargo clippy --all-features --all-targets` passes
- [ ] CI passes


Cargo.toml
```diff
@@ -217,3 +224,38 @@ substrate-bn = { git = "https://github.com/sp1-patches/bn", tag = "patch-0.6.0-s
...
+# The derive macro must match the trait definition version
+alloy-tx-macros = { git = "https://github.com/alloy-rs/alloy", tag = "v1.1.3" }
+
+
+# Note: vergen-lib conflict fixed by pinning vergen=9.0.4, vergen-git2=1.0.5 in Cargo.lock
+
+# Patch reth crates to fix PartialEq bug in reth-transaction-pool (PR #21336)
+# Using succinctlabs fork to enable HTTPS patching in CI (Cargo requires different source URL)
+# v1.9.3-kona branch contains the backported fix from paradigmxyz/reth
+[patch."https://github.com/paradigmxyz/reth"]
```
Farhad-Shabani, line 239: What if we just patch it by pinning `rev = <commit_hash>` and keep using `https://github.com/paradigmxyz/reth` instead?

fakedev9999: IIRC, that won’t work due to a Cargo limitation. When patching a git dependency, the patch source must have a different URL than the original source.

PR #794, opened 2026-02-04
fix(proposer): prevent defense task counter from blocking on skipped games

fakedev9999:
## Problem
When `spawn_game_proving_task` skipped a game due to vkey mismatch or deadline passed, callers still incremented their task counters. This caused a single foreign game to permanently block defense proving if `max_concurrent_defense_tasks` was reached by phantom tasks.

## Solution
Changed `spawn_game_proving_task` to return `Ok(bool)` instead of `Ok(())`, distinguishing between:
- `Ok(true)`: task was spawned (increment counters)
- `Ok(false)`: task was skipped (don't increment)
- `Err`: actual error occurred

The fix ensures `active_defense_tasks_count` and `active_proving` are only incremented when tasks are actually spawned.

## Addresses
https://github.com/succinctlabs/op-succinct/pull/758#pullrequestreview-3747115244

### DIFF
--- a/fault-proof/src/proposer.rs
+++ b/fault-proof/src/proposer.rs
@@ -1840,7 +1840,7 @@ where
 
                     // Spawn proving task
                     match self.spawn_game_proving_task(game_address, false, Some(deadline)).await {
-                        Ok(()) => {
+                        Ok(true) => {
                             tracing::info!(
                                 game_address = ?game_address,
                                 game_index = %index,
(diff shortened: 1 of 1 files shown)


fault-proof/src/proposer.rs
```diff
@@ -1849,6 +1849,7 @@ where
                             spawned_count += 1;
                             active_proving += 1;
                         }
+                        Ok(false) => {}
```
Farhad-Shabani, line 1852: Should log anything here?

Farhad-Shabani (approved):
Just a small comment, otherwise looks good.

fault-proof/src/proposer.rs
fakedev9999: Logging already happens inside `should_skip_proving` depending on the reason. Adding a log here would duplicate those messages.

PR #806, opened 2026-02-17
fix(fault-proof): use l2BlockNumber() for backward compatibility with v1.0.0 games

seolaoh:
## Problem

During the op-contracts v5.0.0 upgrade (#774), the contract function was renamed from `l2BlockNumber()` to `l2SequenceNumber()`. While the Solidity contract maintains `l2BlockNumber()` as a backward-compatible alias, the Rust code was updated to call `l2SequenceNumber()` directly.

This causes a **critical startup failure** during contract upgrades:

1. When upgrading from `OPSuccinctFaultDisputeGame` v1.0.0 to v2.0.0, the `DisputeGameFactory` contains only v1.0.0 games
2. At startup, the proposer/challenger calls `sync_state()` to fetch the v1.0.0 games
3. For v1.0.0 games, calling `l2SequenceNumber()` fails because this function doesn't exist
4. The error causes `sync_state()` to fail and retry in a loop
5. The proposer/challenger never starts successfully

## Solution

Revert to using `l2BlockNumber()` in the Rust code, which exists on:
- ✅ v1.0.0 games (original function)
- ✅ v2.0.0 games (backward compatibility alias)

### DIFF
--- a/fault-proof/src/challenger.rs
+++ b/fault-proof/src/challenger.rs
@@ -354,7 +354,7 @@ where
             return Ok(());
         }
 
-        let l2_block_number = contract.l2SequenceNumber().call().await?;
+        let l2_block_number = contract.l2BlockNumber().call().await?;
         let computed_output_root =
             self.l2_provider.compute_output_root_at_block(l2_block_number).await?;
         let output_root = contract.rootClaim().call().await?;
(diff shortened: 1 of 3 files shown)


fakedev9999:
Thanks for catching this @seolaoh! Great find — the `l2SequenceNumber()` → `l2BlockNumber()` revert is the right fix for backward compatibility during the v1.0.0 → v2.0.0 upgrade.

I picked up your changes in #807 with one additional fix: the test monitor at `fault-proof/tests/common/monitor.rs:134` also needed the same `l2BlockNumber()` switch. Your original authorship is preserved in the commit history.

Closing this in favor of #807 since maintainer push wasn't enabled on the fork. Thanks again for the contribution!

PR #860, opened 2026-04-07
fix(logger): let RUST_LOG override default log suppressions

fakedev9999:
## Summary

- Fix: `RUST_LOG` directives can now override the hardcoded noise-suppression defaults
- Previously, directives like `single_hint_handler=error` were added **after** `RUST_LOG` parsing, always winning — making `RUST_LOG=single_hint_handler=debug` ineffective ([current code](https://github.com/succinctlabs/op-succinct/blob/ca9ca00e8448079b3854fb969ad4ece4fd458a2c/utils/host/src/logger.rs#L15-L33))
- Now defaults are applied first, then `RUST_LOG` directives are layered on top — matching targets get overridden, everything else stays suppressed
- Invalid directives in `RUST_LOG` are logged to stderr and skipped (previously silently ignored)

Based on #770 by @kien-rise — same problem, different fix approach. #770 returns early when `RUST_LOG` is set, which drops all noise suppression. This PR preserves defaults while allowing overrides.

## Test plan

- [x] `cargo check -p op-succinct-host-utils`
- [x] `cargo clippy -- -D warnings`
- [x] `cargo fmt --check`
- [ ] CI

### DIFF
--- a/utils/host/src/logger.rs
+++ b/utils/host/src/logger.rs
@@ -13,11 +13,8 @@ use tracing_subscriber::{
 static INIT: OnceLock<Result<()>> = OnceLock::new();
 
 fn build_env_filter() -> EnvFilter {
-    EnvFilter::try_from_default_env()
-        .unwrap_or_else(|e| {
-            println!("failed to setup env filter: {e:?}");
-            EnvFilter::new("info")
-        })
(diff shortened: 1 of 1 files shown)


kien-rise (approved):
This looks clean and functional.

PR #861, opened 2026-04-07
fix(test): fix race condition in proposer recovery integration test

fakedev9999:
## Summary

- Fix flaky `test_proposer_recovery_after_canonical_head_invalidation` that consistently fails on the eigenda integration variant
- Root cause: single-pass game scan after the first new game appeared — if the proposer hadn't completed recovery yet, the correct game (with expected parent) was missed
- Replace with a polling loop that continuously scans newly created games until the expected parent is found or `WAIT_TIMEOUT` (60s) expires

## Context

This test has been intermittently failing on main since ~4/1, always on the eigenda variant. The eigenda DA layer has slower sync cycles, making the proposer take longer to detect canonical head invalidation and recover — widening the race window.

After this lands, `Fault Proof (integration - eigenda)` can be re-added as a required merge queue check.

## Test plan

- [ ] CI: `Fault Proof (integration - eigenda)` passes
- [ ] CI: `Fault Proof (integration)` still passes (ethereum variant unaffected)


fakedev9999:
/integration

https://github.com/succinctlabs/op-succinct/actions/runs/24078701168

PR #865, opened 2026-04-07
fix(fault-proof): pin L1 block during sync to prevent RPC inconsistency

fakedev9999:
## Summary

With load-balanced RPCs (proxyd, Alchemy, etc.), different `eth_call` requests within the same sync cycle can hit backends at different block heights. This breaks atomicity between related state reads — e.g. `credit()` returns 0 (block N+1) while `anchorGame()` returns the old anchor (block N), causing the game to be incorrectly evicted.

**Root cause of [Celo's March 26 incident](https://github.com/celo-org/op-succinct/issues/132#issuecomment-4162445779):** canonical head cleared, ~470 orphaned games created.

---

### 1. L1 block pinning

Fetch L1 block once at the start of `sync_state()` and thread the pinned block + timestamp through all downstream reads. All state reads within a sync cycle see a consistent L1 snapshot.

**Pinned:** `sync_games` (gameCount, anchorGame, claimData, status, isGameFinalized, gameType, gameCreator, credit), `fetch_game` (gameAtIndex, anchorStateRegistry, l2BlockNumber, rootClaim, wasRespectedGameTypeWhenCreated, status, claimData, vkeys), `sync_anchor_game` (anchorGame), `is_parent_resolved` / `is_parent_challenger_wins` (gameAtIndex, status).

**NOT pinned:** `prove_game`, `resolve_game`, `claim_bond`, `create_game` (tx submission paths). Challenger passes `BlockId::latest()` (separate pinning can follow).

### 2. Confirmation offset (`SYNC_L1_CONFIRMATIONS`)

New env var (default: `0`). Pins reads to `latest - N` for deployments with unstable backends. At default 0, uses the latest block directly from a single RPC response (no split-backend risk). When > 0, fetches the confirmed block separately; if unavailable (`Ok(None)`), the cycle is skipped. `Err` is propagated.

### 3. Skip sync when head unchanged

`last_synced_l1_block` tracks the last successful sync. Skips the cycle when the confirmed block hasn't advanced.

### 4. Backup-restore safety

- **Prune future games** before the status-sync loop — games above `pinned_latest_index` are removed so `compute_canonical_head` never picks a head that doesn't exist at the pinned height.
- **Clear stale anchor** — `sync_anchor_game` clears `state.anchor_game` when the pinned anchor is `Address::ZERO` or not in cache.
- **Reset cursor** when pinned snapshot has zero games — ensures games are rediscovered once confirmed.
- **Preserve anchor baseline** — `canonical_head_l2_block` is intentionally NOT cleared when canonical head index is None, so `should_create_game()` can still propose the first game.

### 5. Duplicate game creation prevention

When `sync_l1_confirmations > 0`, the pinned cache may lag behind tip. Without protection, the proposer could create sibling games with the same parent.

**Guard:** `last_created_game_l2_block` blocks creation when the proposed L2 block ≤ the most recently created game. `last_created_game_address` precisely identifies the guarded game for CHALLENGER_WINS invalidation (globally unique, from receipt event — no race-prone `gameCount()-1`).

- **Persisted** in backup (both fields, `#[serde(default)]` for backward compat)
- **Immediate backup** after successful creation (reduces crash window)
- **No explicit catch-up clear** — guard is naturally bypassed when `next_l2_block > last_created` after canonical head advances
- **CHALLENGER_WINS clear** — guard is reset when the subtree removal contains the exact guarded game (matched by address)

**Residual limitation:** There is still a small crash window if the process dies before the post-creation backup is durably written, or if an earlier backup is in progress and the post-creation backup is skipped. We accept this to avoid synchronous backup latency on the creation path.

### 6. Metrics

New gauges:
- `op_succinct_fp_canonical_head_game_index` (-1 when cleared)
- `op_succinct_fp_anchor_game_index` (-1 when cleared)

Fixed stale metrics: `FinalizedL2BlockNumber` now resets to 0.0 when `get_finalized_l2_block_number` returns `None`. Dead metric reset branch removed (canonical_head_l2_block is no longer cleared).

### 7. RPC call reduction

- Single `get_block_by_number(Latest)` replaces separate `get_block_number()` + `get_block_by_number()`.
- Reuses `anchor_address` from top of `sync_games` instead of re-fetching per game.
- Removed race-prone `gameCount()` call from `create_game` log path.
- Anchor check simplified to `HashMap::contains_key` (O(1) vs O(n)).

---

Addresses [celo-org/op-succinct#132](https://github.com/celo-org/op-succinct/issues/132)
Tracking: [cloud-ops#98](https://github.com/succinctlabs/cloud-ops/issues/98)
Follow-up validation: [cloud-ops#102](https://github.com/succinctlabs/cloud-ops/issues/102), [cloud-ops#103](https://github.com/succinctlabs/cloud-ops/issues/103)

## Test plan

- [x] `cargo check -p op-succinct-fp` / `cargo clippy` / `cargo fmt`
- [x] Backup partial-prune test (game 0 survives, game 1 pruned with confirmations=1)
- [x] Backup zero-games test (all pruned with huge confirmations)
- [x] Backup rediscovery test (prune → mine block → game rediscovered via cursor reset)
- [x] Duplicate creation guard test (should_create_game blocks with stale cache, bypassed after head advance)
- [x] CHALLENGER_WINS guard invalidation test (guard set → game challenged → subtree removed → guard cleared → creation unblocked)
- [x] Backup schema guard updated
- [ ] CI (Fault Proof sync + sysgo e2e)
- [ ] Celo Sepolia deployment for extended testing


fault-proof/src/proposer.rs
```diff
@@ -623,11 +623,23 @@ where
...
+        // Pin L1 block for the entire sync cycle so all state reads see a consistent
+        // snapshot. Without this, load-balanced RPCs can return data from different block
+        // heights, breaking atomicity between related reads (e.g. credit vs anchorGame).
+        // Ref: https://github.com/celo-org/op-succinct/issues/132
+        let l1_block = self
+            .l1_provider
+            .get_block_by_number(BlockNumberOrTag::Latest)
+            .await?
+            .context("Failed to fetch latest L1 block")?;
+        let pinned_block = BlockId::number(l1_block.header.number);
```
palango: I think it would be useful to add a confirmed block configuration, so that the system doesn't relay on the (potentially more unstable) latest block. The default can be low so that it doesn't change the systems behaviour too much.

`confirmed_block = latest_block _ num_confirmations`

fault-proof/src/proposer.rs
```diff
@@ -623,11 +623,23 @@ where
     /// 2. `sync_anchor_game` aligns the cached anchor pointer with the registry contract.
     /// 3. `compute_canonical_head` recomputes the head game used for proposal selection.
     pub async fn sync_state(&self) -> Result<()> {
+        // Pin L1 block for the entire sync cycle so all state reads see a consistent
+        // snapshot. Without this, load-balanced RPCs can return data from different block
+        // heights, breaking atomicity between related reads (e.g. credit vs anchorGame).
+        // Ref: https://github.com/celo-org/op-succinct/issues/132
+        let l1_block = self
```
palango: You could also store the latest synced l1 head in the `OPSuccinctProposer` and skip the sync when the head didn't progress.

palango (commented):
Hi Taehoon,
I had a look at this yesterday as well and some ideas for this PR.

fakedev9999:
> Hi Taehoon, I had a look at this yesterday as well and some ideas for this PR.

Thanks for the review! Implemented both suggestions!

seolaoh (commented):
Could you add metrics for `canonical_head_index` and `anchor_game_index`?                                                                         
                                                                                                                                                                       
Currently the available metric is `canonical_head_l2_block`, which is never cleared when canonical head is set to None (only `canonical_head_index` is reset), so the metric stays stale and the clearing is invisible from the metric side.                                                                  

Also index-based metrics would give us:                                                                                                                                   
- Canonical head clear detection: `canonical_head_index` dropping to 0/-1 is a direct signal for alerting
- Orphan game detection: a decrease in `canonical_head_index` catches the orphan games we've been seeing
- Anchor/head correlation: overlaying both on the same graph makes it easy to spot when they converge (regression) vs diverge (healthy gap)

fault-proof/src/proposer.rs
```diff
@@ -623,15 +627,62 @@ where
...
+        // snapshot. Without this, load-balanced RPCs can return data from different block
+        // heights, breaking atomicity between related reads (e.g. credit vs anchorGame).
+        // Ref: https://github.com/celo-org/op-succinct/issues/132
+        let latest_block = self
+            .l1_provider
+            .get_block_by_number(BlockNumberOrTag::Latest)
+            .await?
+            .context("Failed to fetch latest L1 block")?;
+        let confirmed_number =
+            latest_block.header.number.saturating_sub(self.config.sync_l1_confirmations);
```
seolaoh, line 654: One concern with `sync_l1_confirmations > 0`: since `should_create_game` reads `canonical_head_l2_block` from the proposer state (which is updated from the pinned sync), the canonical head won't advance until the pinned block sees the newly created game.

With `sync_l1_confirmations = 10` (~120s at 12s/block) and `fetch_interval = 30s`, the proposer will call `should_create_game` ~4 times before the pinned sync catches up to the new game. Each time, it gets the same `(next_l2_block, parent_game_index)`. The `handle_game_creation` while loop then finds the existing game at `latest`, increments l2_block by 1, and creates a **duplicate sibling** — orphaning the original.

We've already observed this pattern on mainnet even without this PR, when L1 RPC instability caused tx confirmation timeouts — the proposer retried and created siblings (e.g. game 31527 was orphaned this way). With a large `sync_l1_confirmations`, this becomes systematic rather than occasional.

Note: we can't simply skip creation when a game exists at the target l2_block — the while loop is necessary for the case where an existing game was resolved as CHALLENGER_WINS and needs a valid replacement.

A possible fix: compare game count at `latest` vs at the pinned block before spawning creation:

```rust
// In should_create_game, before returning true:
let latest_count = self.factory.gameCount().call().await?;
let pinned_count = self.factory.gameCount().block(pinned_block).call().await?;
if latest_count > pinned_count {
    // Games exist at tip that pinned sync hasn't seen yet — likely ours.
    // Skip creation until pinned sync catches up.
    return Ok((false, ...));
}
```

fakedev9999: This is addressed in this PR. The last_created_game_l2_block guard in should_create_game() handles exactly this case.

After `handle_game_creation()` succeeds, it records the created L2 block. On subsequent loops where the pinned cache still shows the same stale canonical head, `should_create_game()` detects `next_l2_block <= last_created` and returns false, skipping creation entirely until the cache catches up.

The guard is also persisted in the backup file so it survives restarts, and cleared in `sync_state()` once canonical_head_l2_block advances past the created game.

We chose this over the `latest_count > pinned_count` approach because game count comparison is fragile with mixed game types, foreign games, and concurrent factory usage. The L2-block guard is scoped to our own creation decisions only.

fakedev9999:
> Could you add metrics for `canonical_head_index` and `anchor_game_index`?
> 
> Currently the available metric is `canonical_head_l2_block`, which is never cleared when canonical head is set to None (only `canonical_head_index` is reset), so the metric stays stale and the clearing is invisible from the metric side.
> 
> Also index-based metrics would give us:
> 
> * Canonical head clear detection: `canonical_head_index` dropping to 0/-1 is a direct signal for alerting
> * Orphan game detection: a decrease in `canonical_head_index` catches the orphan games we've been seeing
> * Anchor/head correlation: overlaying both on the same graph makes it easy to spot when they converge (regression) vs diverge (healthy gap)

Added index-based proposer gauges:

- `op_succinct_fp_canonical_head_game_index`
- `op_succinct_fp_anchor_game_index`

They use `-1` as the absent/cleared sentinel because game index `0` is valid.

One important semantic note: I did not clear the underlying `canonical_head_l2_block` state when the cached canonical head is absent. That field is also used as the anchor-derived baseline for first-game creation; clearing it blocks fresh deployments / zero-game pinned snapshots from proposing. So the canonical-head clear signal is now the index gauge dropping to `-1`, while the L2-block gauge may continue to expose the preserved proposal baseline.

Also fixed `op_succinct_fp_finalized_l2_block_number` so it resets to `0` when
finalized lookup returns `None`, instead of keeping a stale finalized value.

fault-proof/src/proposer.rs
fakedev9999: Addressed this without using the raw `latest_count > pinned_count` gate.

The new behavior is:
- after successful game creation, the proposer records `last_created_game_l2_block`
- `should_create_game()` now skips creation while the pinned cache would propose at or before that recently-created L2 block
- the guard is cleared once pinned sync catches up and canonical head reaches that L2 block
- the guard is persisted in proposer backups and restored on restart
- we trigger a backup immediately after successful creation to reduce the restart window

I kept the existing `handle_game_creation()` while-loop intact so replacements can still be created when the target block is occupied by a game that should no longer be extended.

I also avoided the broad `gameCount(latest) > gameCount(pinned)` gate because it can include unrelated factory games; the guard is scoped to the proposer’s own successful creation path.

Residual limitation is documented in the PR: there is still a narrow crash/upgrade window before the post--creation backup is durably available, and legacy backups do not contain the new guard field.

fault-proof/src/proposer.rs
```diff
@@ -1570,7 +1694,10 @@ where
                 ProposerGauge::AnchorGameL2BlockNumber.set(0.0);
             }
         } else {
-            tracing::warn!("canonical_head_l2_block is None; skipping metrics update");
+            // Reset L2 block metrics when canonical head is cleared to avoid stale values.
```
seolaoh: If we don't clear `canonical_head_l2_block` when canonical head is cleared, do we need this block?

fault-proof/src/proposer.rs
seolaoh: Makes sense, thanks for the detailed response!

fault-proof/src/proposer.rs
```diff
@@ -650,15 +717,56 @@ where
...
+                }
+                tracing::info!(
+                    count = future_games.len(),
+                    "Pruned games above pinned latest index"
+                );
+                // Clear anchor if it pointed to a pruned game.
+                let should_clear_anchor = state
+                    .anchor_game
+                    .as_ref()
+                    .is_some_and(|a| !state.games.values().any(|g| g.address == a.address));
```
piersy: Nit: Could be simplified to hashmap lookup.
```suggestion
                    .is_some_and(|a| !state.games.contains_key(&a.index));
```

fault-proof/src/proposer.rs
fakedev9999: @seolaoh Good catch. Since we preserve canonical_head_l2_block now (anchor baseline), the else branch is unreachable. Removed the dead metric resets, kept the warn log as a defensive signal.

fault-proof/src/proposer.rs
fakedev9999: @piersy Thanks! Applied contains_key(&a.index). Since it's is O(1) vs the O(n) value scan.

fault-proof/src/proposer.rs
```diff
@@ -623,15 +634,71 @@ where
...
 
+        self.last_synced_l1_block.store(confirmed_number, Ordering::Relaxed);
+
+        // Clear the creation guard once the cache has caught up to the created game.
+        let last_created = self.last_created_game_l2_block.load(Ordering::Relaxed);
+        if last_created > 0 {
+            let head_l2 = self.state.read().await.canonical_head_l2_block;
+            if head_l2.is_some_and(|h| h.to::<u64>() >= last_created) {
+                self.last_created_game_l2_block.store(0, Ordering::Relaxed);
+            }
```
piersy: Hey @fakedev9999 I'm wondering why you clear `last_created_game_l2_block` here. The backup is persisted just after returning from `sync_state`, so if the proposer restarts then and gets a lower pinned block it could end up making a duplicate game (since the guard has been disabled).

It seems like just removing this block would fix that. Maybe I'm missing something but I can't see any value in this block.

fault-proof/src/proposer.rs
piersy, line 1008: When a subtree is removed due to CHALLENGER_WINS I think we need to reset `last_created_game_l2_block`.
```suggestion
                    GameSyncAction::RemoveSubtree(index) => {
                        // Reset the duplicate-creation guard if the subtree being
                        // removed contains the game it was tracking. Without this,
                        // the guard would permanently block creation after a
                        // legitimate branch switch caused by CHALLENGER_WINS.
                        let last_created =
                            self.last_created_game_l2_block.load(Ordering::Relaxed);
                        if last_created > 0 {
                            let subtree = state.descendants_of(index);
                            let guard_invalidated = subtree.iter().any(|idx| {
                                state
                                    .games
                                    .get(idx)
                                    .is_some_and(|g| g.l2_block.to::<u64>() == last_created)
                            });
                            if guard_invalidated {
                                self.last_created_game_l2_block
                                    .store(0, Ordering::Relaxed);
                                tracing::info!(
                                    last_created,
                                    root_index = %index,
                                    "Reset creation guard: tracked game removed by CHALLENGER_WINS"
                                );
                            }
                        }

```

fault-proof/src/proposer.rs
fakedev9999: Thanks! I removed the explicit catch-up clear. Once canonical head advances, `should_create_game()` naturally bypasses the guard because `next_l2_block > last_created`, so explicitly clearing it was unnecessary and could re-open the restart case by persisting `0`.

fault-proof/src/proposer.rs
fakedev9999: Applied, with one refinement: I switched the targeted invalidation from game index to game address. The index path depended on `gameCount() - 1`, which is race-prone under concurrent creation, while the address comes directly from the `DisputeGameCreated` receipt event and is globally unique.

I also added a regression test (`test_challenger_wins_clears_creation_guard`) covering the full path: guard set -> guarded game removed by `CHALLENGER_WINS` subtree removal -> guard cleared -> creation unblocked.

fault-proof/src/proposer.rs
```diff
@@ -270,6 +270,15 @@ where
...
     pub identity: ProposerIdentity,
+    /// L1 block number used in the last successful sync cycle. Sync is skipped when the
+    /// pinned block hasn't advanced past this value.
+    last_synced_l1_block: Arc<AtomicU64>,
+    /// L2 block number of the most recently created game. Used to prevent duplicate
+    /// game creation when the pinned sync cache lags behind the chain tip.
+    last_created_game_l2_block: Arc<AtomicU64>,
+    /// Address of the most recently created game. Used to precisely identify
+    /// the guarded game for CHALLENGER_WINS subtree removal.
+    last_created_game_address: Arc<Mutex<Address>>,
```
Farhad-Shabani, line 281: Could we place these under the `ProposalState`?


And as for `last_create_game_***`, since both fields are already persisted in `ProposerBackup`, maybe for an enhanced clarity:

```rust
pub struct CreatedGameRef {
     pub l2_block: u64,
     pub address:  Address,
 }

pub struct ProposerState {
     ...
     pub last_created_game: CreatedGameRef,
 }
```

fault-proof/src/proposer.rs
```diff
@@ -623,15 +641,62 @@ where
...
+                        "Confirmed block not available on this backend, skipping sync cycle"
+                    );
+                    return Ok(());
+                }
+            }
+        };
+
         // Pull new games and synchronize cached game statuses.
-        self.sync_games().await?;
+        self.sync_games(pinned_block, pinned_timestamp).await?;
```
Farhad-Shabani, line 690: Quick suggestion:
`BlockId` rides through a bunch of signatures. Wouldn't a small wrapper here go long? moving the decision into the type system? Something like:

```rust
pub struct PinnedView<'a, P> {
   factory:  &'a FactoryContract<P>,
    registry: &'a AnchorRegistryContract<P>,
    block:    BlockId,
   timestamp: u64,
 }

impl<'a, P: Provider> PinnedView<'a, P> {
   pub async fn game_count(&self) -> Result<U256> {
        self.factory.gameCount().block(self.block).call().await.map_err(Into::into)
   }
   pub async fn game_at_index(&self, idx: U256) -> Result<GameView<'_, P>> { ... }
   pub async fn anchor_game(&self) -> Result<Address> { ... }
   pub fn timestamp(&self) -> u64 { self.timestamp }
}
```

fault-proof/src/proposer.rs
fakedev9999: Good suggestion. Grouping these into a `CreatedGameRef` would be cleaner.

I'd prefer to keep that as a follow-up though: the guard fields are currently kept outside the main `ProposerState` lock on purpose, so `handle_game_creation()` and `should_create_game()` can update/check them without going through the `RwLock<ProposerState>`. Moving them into `ProposerState` is doable, but it changes the state/locking shape for a bug-fix PR that is otherwise already functionally settled.

I'd rather keep this PR focused on the correctness fixes and take the struct/state-layout cleanup separately if we want it.

fault-proof/src/proposer.rs
fakedev9999: I considered wrapping the pinned reads, but for this PR I chose to keep the `BlockId` threading explicit.

At the moment the surface area is still fairly small, and passing `BlockId` directly keeps the pinned/read behavior easy to trace in the sync path. A `PinnedView` wrapper would be a reasonable cleanup if this grows further, but right now it felt like adding another abstraction layer without reducing enough complexity to justify it.

If pinning expands beyond the current sync/read path, I'm happy to revisit that in a follow-up.

fault-proof/src/proposer.rs
Farhad-Shabani: Sounds good. I’ll leave that with you if you’d like to handle it in a follow-up PR.

fault-proof/src/proposer.rs
Farhad-Shabani: Same comment: https://github.com/succinctlabs/op-succinct/pull/865#discussion_r3115000960

Farhad-Shabani (approved):
✨

fault-proof/src/proposer.rs
fakedev9999: Thanks! Tracked in https://github.com/succinctlabs/apps-ops/issues/7 for post-merge cleanup.

PR #871, opened 2026-04-12
fix(fault-proof): correct resolve() invariant comment

cuiweixie:
## Summary
Corrects the `resolve()` INVARIANT comment in `OPSuccinctFaultDisputeGame`: it previously said resolution cannot occur *unless* the game was already resolved, which contradicts the check (`status != IN_PROGRESS` → revert when not in progress). The intended meaning is that resolution cannot occur *if* the game has already been resolved.

## Test plan
- [ ] N/A (comment-only change)


fakedev9999:
Thanks for the fix! Cherry-picked as #881 (with you as co-author) so CI can run under our auth. Closing in favor of that PR.

PR #882, opened 2026-04-16
feat: configurable L1 block selection for proposer

fakedev9999:
## Summary

Adds `L1_BLOCK_TAG` (`finalized` | `safe` | `latest`) and `L1_CONFIRMATIONS` env vars so operators can choose which L1 block the proposer anchors proof generation against, instead of being hardcoded to `finalized` (~13 min).

Default is `finalized` / `0`, preserving byte-identical behavior across all DA backends.

Motivation: Polygon/Agglayer requested non-default selection to enable faster bridging. The new config is generic and aligns with EF's [Fast Confirmation Rule](https://fastconfirm.it/), which reuses the `safe` tag — once FCR ships in consensus clients, `safe` automatically tightens from ~6.4 min to ~13 sec without further code changes.

## Architecture

- `L1BlockSelectionConfig` is owned by `OPSuccinctDataFetcher` (single source of truth). Parsed via:
  - `from_env()` at proposer entry points → clean startup error on invalid values
  - `from_env_or_default()` inside library constructors → panic with env var context for utility scripts
- ETH/EigenDA hosts use a default/non-default match arm so the default path keeps the original `BlockId::finalized()` call sites byte-for-byte. Non-default resolves the max provable L2 block via `optimism_safeHeadAtL1Block(resolved_l1_number)`.
- Celestia is Blobstream-driven and does not honor the knob; rejected at startup via `OPSuccinctHost::supports_non_default_l1_selection`.
- Shared `enforce_l1_selection_supported` helper invoked uniformly by proposer binaries and operator-facing utility scripts.

## Safety

- **ETH/EigenDA**: non-default requires SafeDB on op-node (proposer hard-fails at startup otherwise). `SAFE_DB_FALLBACK` does not apply to the non-default L1 → L2 resolution path.
- **Celestia**: non-default rejected at startup across production proposer binaries and covered operator-facing utility scripts.
- `get_rolling_block_range`: now returns explicit error on `range > end` (was raw subtraction → underflow).
- Search-hint subtractions in `block_range` use `saturating_sub`.
- `split_range_based_on_safe_heads` takes `&OPSuccinctDataFetcher` to avoid bypassing the configured fetcher with fresh `default()` instances.

## Entrypoints covered (Celestia rejection enforced)

| Entrypoint | Type |
|---|---|
| `validity/bin/validity.rs` | Production proposer |
| `fault-proof/bin/proposer.rs` | Production proposer |
| `scripts/prove/bin/multi.rs` | Utility (local proving) |
| `scripts/utils/bin/cost_estimator.rs` | Utility (cost estimation) |
| `scripts/utils/bin/fetch_l2oo_config.rs` | Utility (setup) |
| `scripts/utils/bin/fetch_fault_dispute_game_config.rs` | Utility (setup) |
| `scripts/utils/bin/gen_sp1_test_artifacts.rs` | Utility |
| `scripts/utils/bin/preflight.rs` | Utility |

Test harnesses (`scripts/prove/tests/*`) and internal tools that construct fetcher/host directly are intentionally out of scope.

## Tests

16 new unit tests in `op-succinct-host-utils`:
- Config parsing & env validation (default, parse, case-insensitive, invalid tag, non-numeric confirmations)
- Panic message includes env var name + offending value
- `rolling_range_subtraction_is_checked` (underflow guard)
- `reference_subtraction_saturates_below_two_hours` (search-hint guard)
- `split_range_basic` invariants (3 cases)
- `to_block_id_matches_tag`

## Test plan

- [ ] `cargo check --workspace --all-targets` — pass locally
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — pass locally
- [ ] `cargo test --workspace --lib` — all green locally (16 new + pre-existing)
- [ ] CI pipeline
- [ ] Smoke test against an ETH-DA proposer with default config (byte-identical behavior)
- [ ] Smoke test against an ETH-DA proposer with `L1_BLOCK_TAG=safe`
- [ ] Verify Celestia build hard-fails at startup with non-default selection


fakedev9999:
Reopening from upstream branch (#883) so CI gets repo secrets (`L1_RPC`, `L2_RPC`, etc.). Fork PRs don't inherit those, which caused `tests` and `cost-estimator` to fail.

PR #883, opened 2026-04-16
feat: configurable L1 block selection for proposer

fakedev9999:
This PR makes the proposer’s L1 anchor configurable via `L1_BLOCK_TAG` (`finalized` / `safe` / `latest`) and `L1_CONFIRMATIONS`.

Default remains `finalized` / `0`, so the default proving path keeps historical behavior.

The motivation is faster bridging: operators can opt into a less conservative L1 anchor when they want lower latency, while keeping the current finalized-based behavior as the default.

A few important semantics:

- On Ethereum/EigenDA, non-default selection uses `optimism_safeHeadAtL1Block(...)` to derive the max provable L2 block, so SafeDB is required.
- On Celestia, non-default selection is rejected on covered entrypoints, since the proving path is Blobstream-driven and the knob would otherwise be a silent no-op.
- The bootstrap config scripts no longer inherit non-default proposer selection when deriving their default `STARTING_*_BLOCK_NUMBER`; their default path stays rooted at literal L2 finalized unless explicitly overridden.

As part of this cleanup, `get_finalized_l2_block_number` was renamed to `get_max_provable_l2_block_number`, and metrics now distinguish literal finalized L2 from max provable L2 instead of overloading one metric with both meanings.

Validated with targeted local fmt/check/test runs plus the PR CI suite.


validity/bin/validity.rs
Farhad-Shabani, line 1: ([validity/src/proposer.rs:1715](https://github.com/succinctlabs/op-succinct/blob/3fd76b8a68fea6e506e1e5d92e81c1eb64992f87/validity/src/proposer.rs#L1715)) L2FinalizedBlock reads BlockId::finalized() directly, so under non-default it diverges from the actual proof anchor. Should relabel or add a separate gauge?

utils/host/src/block_range.rs
Farhad-Shabani: Should be updated. No longer true under non-default.

utils/host/src/block_range.rs
```diff
@@ -178,3 +192,80 @@ pub async fn split_range_based_on_safe_heads(
 
     Ok(ranges)
 }
+
+#[cfg(test)]
+mod tests {
```
Farhad-Shabani, line 200: Should we drop `rolling_range_subtraction_is_checked` and `reference_subtraction_saturates_below_two_hours`? they're asserting against `u64::checked_sub` / `saturating_sub `directly, so they're really testing stdlib.

utils/celestia/host/src/host.rs
Farhad-Shabani: should we rename `get_finalized_l2_block_number`? something like `get_max_provable_l2_block` ?
Cause under non-default it returns the L2 head at the configured L1 tag.

Farhad-Shabani, line 71: Same comment: https://github.com/succinctlabs/op-succinct/pull/883/changes#r3175537843

scripts/utils/bin/fetch_fault_dispute_game_config.rs
Farhad-Shabani: Note that under e.g. `L1_BLOCK_TAG=safe` this Err branch dispute game's `STARTING_L2_BLOCK_NUMBER` ends up rooted at a less-final anchor than under default.
Basically the starting L2 block here is bootstrap config, thus conflating the two means an operator who turns on safe for proposer latency also shifts their dispute-game bootstrap to a reorg-prone anchor, probably without realizing it. Probably fine though?

Same applies to `scripts/utils/bin/fetch_l2oo_config.rs`

Farhad-Shabani (changes requested):
Nice addition, thanks! 
Asked for a few small touches, but otherwise looks great.

validity/bin/validity.rs
fakedev9999: Done. Added `succinct_l2_max_provable_block` and kept `succinct_l2_finalized_block` literal. Did the same split on the fault-proof side for consistency.

utils/host/src/block_range.rs
fakedev9999: Updated to say host-resolved max provable, not finalized.

utils/host/src/block_range.rs
fakedev9999: Dropped both. Agreed they were just testing stdlib behavior.

utils/celestia/host/src/host.rs
fakedev9999: Renamed to `get_max_provable_l2_block_number` across trait, impls, call sites, and docs.

scripts/utils/bin/fetch_fault_dispute_game_config.rs
fakedev9999: Fixed in both scripts. The default bootstrap path now forces `L1BlockSelectionConfig::default()`, so `STARTING_*_BLOCK_NUMBER` stays rooted at literal L2 finalized unless explicitly overridden.

Farhad-Shabani (approved):
Great job! Thanks for addressing the comments!

PR #917, opened 2026-05-26
fix: remove yanked core2 dependency and refresh Celestia ELF

fakedev9999:
## Summary
- Removes the yanked `core2` crate from the proof dependency path by bumping `cid`/`multihash` where possible and vendoring a patched `multihash 0.16.3` using `no_std_io2` for the `ethereum-consensus` path.
- Fixes the attached CI failures by excluding vendored patch Markdown from lychee and pinning `sqlx-cli` to `0.8.6` for the repo toolchain.
- Rebuilds `elf/celestia-range-elf-embedded` on `succinct-gpu-06` with SP1 `v6.1.0` so the committed ELF matches the patched dependency graph.
- Widens Celestia's rolling max-provable-block search window while preserving the default 3,600-block window for other DA backends, avoiding Blobstream/indexer lag failures in `cycle-count-diff-celestia`.

## Verification
- `cargo fmt --all`
- `cargo clippy --all-features --all-targets -- -D warnings -A incomplete-features`
- `cargo install sqlx-cli --version 0.8.6 --locked --no-default-features --features native-tls,postgres`
- `cd validity && cargo sqlx prepare --check`
- GitHub `check_links` passed after the lychee workflow fix.
- GitHub `Verify SQLx queries` passed after the SQLx CLI pin.
- GitHub `elf` passed after rebuilding the Celestia ELF.
- GitHub `tests` passed on rerun; the first failure was an embedded PostgreSQL startup flake.
- Static ELF checks for `celestia-range-elf-embedded`: RISC-V ELF format, patched bn/k256/p256 paths, 40 ECALL sites via `llvm-objdump-14` on GPU 06, cycle-tracker strings, and `OpZkvmPrecompiles` symbol.
- Rebuilt Celestia ELF sha256: `2ea35ec30755894e210cd76627dee2298dc240d972333cd07599b7ac821e6989`.


fakedev9999:
Superseded by #920, which uses the merged upstream ethereum-consensus fix and refreshes the Celestia ELF without the temporary vendored multihash patch.

PR #923, opened 2026-05-29
fix(validity): anchor agg checkpoint to safe L1 head to avoid tip reorg

Farhad-Shabani:
### Description
The proposer checkpoints an L1 block (`checkpointBlockHash`) that the aggregation guest later proves against — but it checkpointed the **unfinalized tip** (`BlockId::latest()`). The checkpoint head is pinned **by hash**; the guest's header range is fetched **by number** (`get_header_preimages`). A tip reorg between the two orphans the checkpoint head, so the guest's `assert_eq!` (`programs/aggregation/src/main.rs:57`) rejects the input and the request goes `UNEXECUTABLE`. Hit 3× on Mantle; each wastes one aggregation request.

### DIFF
--- a/validity/.sqlx/query-4638ae1172f5f6e5a55633747989fd5ddabe6f9b613578cf18e927dd6a01bf90.json
+++ b/validity/.sqlx/query-4638ae1172f5f6e5a55633747989fd5ddabe6f9b613578cf18e927dd6a01bf90.json
@@ -1,0 +1,29 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT MAX(l1_head_block_number) AS max_l1_head FROM requests WHERE range_vkey_commitment = $1 AND rollup_config_hash = $2 AND status = $3 AND req_type = $4 AND start_block >= $5 AND end_block <= $6 AND l1_chain_id = $7 AND l2_chain_id = $8",
+  "describe": {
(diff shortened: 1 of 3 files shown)


fakedev9999:
@Farhad-Shabani I think a cleaner direction is to anchor the checkpoint to the batch's max l1Head instead of any fresh tag? Wdyt?

Farhad-Shabani:
> I think a cleaner direction is to anchor the checkpoint to the batch's max l1Head instead of any fresh tag? Wdyt?

Agreed! That's more fundamental fix, as the guest only needs checkpoint >= max(l1Head) across the batch's range proofs. Anchored to the batch's max `l1Head`, read from the persisted `l1_head_block_number` column (`MAX` over the same range-proof set the agg consumes.

Two things to call out:
1. I checkpoint `max(safe, batch_max_l1_head)`. Plain safe would be wrong under `L1_BLOCK_TAG=latest`.
2. The floor now also gates checkpoint reuse. The reuse path validated the cached checkpoint only against on-chain `historicBlockHashes`.

Caveat I left as a known limitation: under `latest`, a boot `l1Head` above `safe` that gets orphaned is reorg-exposed regardless of checkpoint choice, which is inherent to `latest`.

fakedev9999 (approved):
lgtm!

PR #938, opened 2026-07-02
chore(deps): pin kona proof read fix

fakedev9999:
## Summary
- Pins optimism/kona monorepo dependencies to `fakedev9999/optimism@17af43b1a158266fa8357c0a7cf11a2fb50a6c82`.
- Pulls in fakedev9999/optimism#4, which keeps v1.6.0 reset-walk semantics but avoids hydrating the full L2 transaction trie when `l2_block_info_by_number` only needs tx0.
- Draft because this depends on a fork commit; once upstreamed, replace the fork rev with the upstream tag/commit.

## Measurement
Katana range: `36207424~36207429`

| Version | Witness preimage bytes | Delta vs v1.6.0 |
| --- | ---: | ---: |
| pre-#928 base | 18,479,023 | - |
| current main / kona v1.6.0 | 20,120,288 | - |
| patched fork rev | 19,243,741 | -876,547 |

The patch recovers 876,547 bytes, or 53.41% of the measured +1,641,265-byte regression on this range. The savings concentrate in pipeline construction: `1,651,074 -> 774,527` bytes. Host inputs, payload derivation, and block execution were unchanged in this measurement.

## Verification
- `cargo check -p op-succinct-prove --tests`
- op-succinct host witness measurement on Katana range `36207424~36207429`


fakedev9999:
Superseded by #939, which uses a branch in succinctlabs/op-succinct so the existing checkout workflows can fetch github.head_ref from the base repository.

PR #939, opened 2026-07-02
perf: reduce oracle verify cycles after kona bump

fakedev9999:
## Summary

This PR restores oracle-verify performance after the kona v1.6.0 bump by combining two targeted fixes:

- pin kona to `fakedev9999/optimism@654eb009c7ff6874fe92192e4771a72f9d508273`, which limits the L1BlockInfo reset walkback to the header plus transaction-0 proof instead of hydrating the full transaction trie for each walked-back block
- specialize witness Keccak validation in the zkVM to call the SP1 Keccak permutation syscall directly, avoiding the slower generic digest/alloy path while preserving the same preimage-key check

## Measurement

Pinned Katana range: `36257802~36257807` (5 blocks, 13 txs, 2,414,892 gas).

| version | oracle-verify | total instruction count | total SP1 gas |
| --- | ---: | ---: | ---: |
| pre-#928 | 111,733,165 | 1,026,784,397 | 1,434,135,661 |
| current main post-#928 | 164,816,721 | 1,192,499,782 | 1,620,109,849 |
| kona reset-walk fix only | 153,461,261 | 1,044,716,993 | 1,450,407,560 |
| this PR | 103,157,045 | 994,359,084 | 1,408,452,516 |

Witness bytes on the same range:

| version | preimages | bytes | Keccak bytes |
| --- | ---: | ---: | ---: |
| pre-#928 | 36,319 | 25,094,051 | 25,092,525 |
| current main post-#928 | 39,920 | 26,851,472 | 26,850,056 |
| kona reset-walk fix | 36,595 | 25,171,067 | 25,169,651 |

## Correctness / safety

The witness predicate is unchanged: Keccak preimages are still accepted only when `PreimageKey::new(expected_hash, key.key_type())` matches the provided key. The zkVM path only replaces how the expected Keccak hash is computed.

Validation run locally:

- `cargo test -p op-succinct-client-utils check_preimage -- --nocapture`
- `cargo fmt --check -- utils/client/src/witness/preimage_store.rs`
- rebuilt range ELFs with `just build-range-elfs` and aggregation with `just build-agg-elf`
- reran pinned Katana `cycle_count_diff` with canonical RPC env loaded without printing values

I also tested switching alloy-primitives to `tiny-keccak`; that made oracle-verify worse (`214,586,279`), so this PR keeps `sha3-keccak` and specializes only this hot witness-validation path.


fakedev9999:
CI note: required checks are green. The remaining red check is the non-required Celestia cycle diff. I reran it once and it failed the same way before producing any perf comparison: `utils/host/src/block_range.rs:92` panicked because the Celestia host returned no max provable L2 block for the rolling range. Normal cycle diff and EigenDA cycle diff both passed, and ELF/tests/lint are green. This looks like the existing Celestia rolling range/indexer availability issue, not this PR's Keccak/kona perf change.

PR #948, opened 2026-07-16
backport(bindings): preserve generated bytecode (#947)

fakedev9999:
## What
- Backport #947 to build full contract artifacts before generating Rust bindings.

## Why
- Preserve bytecode constants consumed by fault-proof tests with current Foundry versions.

## Validation
- `PATH="/private/tmp/foundry-current/bin:$HOME/.sp1/bin:$PATH" cargo fmt --all -- --check`
- `PATH="/private/tmp/foundry-current/bin:$HOME/.sp1/bin:$PATH" cargo check -p op-succinct-fp --tests`
- `PATH="/private/tmp/foundry-current/bin:$HOME/.sp1/bin:$PATH" cargo check --all-targets --all-features --tests`
- Confirmed ELF hashes are unchanged.

### DIFF
--- a/bindings/build.rs
+++ b/bindings/build.rs
@@ -27,6 +27,14 @@ fn main() -> anyhow::Result<()> {
         return Ok(());
     }
 
+    // Generate full artifacts because the bindings expose contract bytecode.
+    let status =
+        Command::new("forge").arg("build").current_dir(&contracts_package_path).status()?;
+
+    if !status.success() {
+        anyhow::bail!("Forge build failed with exit code: {}", status);
+    }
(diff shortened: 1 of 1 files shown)


fakedev9999:
CI note: the Lite Docker jobs fail before compilation on the redundant v3.x rustup reinstall. #949 backports the existing fix, and all four Lite image jobs pass there. Conversely, #949 lint hits the generated `BYTECODE` issue fixed by this PR. The changes remain independent and source-equivalent, so this PR should land first; #949 can then rebase and rerun.

PR #953, opened 2026-07-24
test(eigenda): trace DA host timeout

fakedev9999:
## What
- Enable Kona hint tracing in the EigenDA witness integration test.
- Fail witness generation after 180 seconds so CI preserves a focused error instead of hitting the one-hour job timeout.

## Why
Identify the fine-grained preimage hint behind the repeated JSON-RPC -32602 response on main commit 13716c2.

## Validation
- cargo fmt --all -- --check
- cargo test -p op-succinct-eigenda-host-utils --features integration --release --no-run

## Notes
Temporary RCA instrumentation; do not merge.

### DIFF
--- a/utils/eigenda/host/tests/witness_generator.rs
+++ b/utils/eigenda/host/tests/witness_generator.rs
@@ -85,7 +85,6 @@ mod integration {
     use alloy_rpc_client::RpcClient;
     use anyhow::{Context, Result};
     use op_succinct_host_utils::fetcher::OPSuccinctDataFetcher;
-    use tracing::{error, info};
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_failing_raw_header_probe() -> Result<()> {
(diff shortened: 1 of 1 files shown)


fakedev9999:
RCA result (temporary diagnostics only):

- Reproduced on L2 blocks 8497795 -> 8497796.
- The failing hint is `l2-block-header 0x01ec1aa2801be61f3e62032e6990cdffe6ca0df3e63ea5e736d85ce8a9da2113`.
- Kona maps that hint to `debug_getRawHeader(hash)`. The RPC returns `-32602 Invalid params: number too large to fit in target type` and Kona retries without backoff or a terminal error.
- The test timeout cancels `host.run`, but its spawned server task is detached on cancellation, so the process still does not exit; I force-cancelled the diagnostic run after preserving logs.

Diagnostic job: https://github.com/succinctlabs/op-succinct/actions/runs/30107854502/job/89530378549

This PR remains a do-not-merge diagnostic.

fakedev9999:
Exact CI-secret probe result:

- `eth_getBlockByNumber(8497788)`: succeeds, hash `0x01ec1aa2…9da2113`.
- `eth_getBlockByHash(0x01ec1aa2…9da2113)`: succeeds and returns block 8497788.
- `debug_getRawHeader(8497788)`: succeeds, 629 bytes, keccak `0x01ec1aa2…9da2113`.
- `debug_getRawHeader(0x01ec1aa2…9da2113)`: fails with `-32602 Invalid params: number too large to fit in target type at line 1 column 67`.

Conclusion: this L2 RPC backend parses `debug_getRawHeader` input as a block-number quantity rather than accepting Kona’s hash parameter. It tries to deserialize the 256-bit hash into its numeric target type and overflows. Kona then retries that deterministic incompatibility forever. Temporary diagnostic; do not merge.

PR #968, opened 2026-08-18
feat(validity): optional gRPC server for externally driven aggregation

hadjiszs:
## Context

The agglayer maintains a very light fork of `op-succinct`. The divergence is only on the client side - in particular no changes to any programs.

This PR brings that light diff upstream, mainly @vcastellm's work, so that our light fork can be archived as soon as this PR get merged.

## Changes

Adds an optional `agglayer` cargo feature exposing a gRPC service (`validity/proto/proofs.proto`) so an external coordinator can decide when to aggregate: it picks the block range, asks for the proof to consume it downstream. `GRPC_ADDRESS` is required when the feature is on. Default builds are unaffected and still need no `protoc`.

Also here:
- `AGG_PROOF_MODE` accepts `compressed`, for proofs consumed by another proof.
- `insert_request` returns the new row id, so `GetMockProof` can read a stored mock proof back. Existing callers ignore it.
- `Proposer::run` takes `Arc<Self>`, to share it with the server task.

Images `op-succinct-agglayer{,-altda}` come from `validity/Dockerfile.agglayer{,.altda}`.


hadjiszs (commented):
I'm good to close it in favor of #980

fakedev9999:
Superseded by #980

PR #969, opened 2026-08-25
ci: validate SP1 version before Docker builds

fakedev9999:
## Summary

- Check that every Dockerfile uses the SP1 SDK version from `Cargo.toml`.
- Stop the Docker workflow before image publication when a version differs.

## Motivation

The existing lint check runs separately from Docker publication.
This check applies the version rule in the publication workflow.

## Validation

- The check passes with the current Dockerfiles.
- The check fails when a Dockerfile uses another SP1 version.
- The workflow YAML parses successfully.


fakedev9999:
Replaced by #970, which uses an upstream branch.

Issue #981, opened 2026-09-02
[Challenger] Make game validation fail-closed and retryable across sync cycles


KyrinCode:
# [Challenger] Make game validation fail-closed and retryable across sync cycles

## Summary

The fault-proof challenger currently has several failure modes where an invalid claim may remain
unchallenged, or where a single unobservable game can prevent the challenger from processing other
games.

This is especially important because the dispute game does not require a proposer to prove that
the claimed L2 height or root is valid at game creation time. If no challenge is submitted before
the challenge window expires, an invalid game may win without a proof.

### 1. Untrusted L2 block numbers can panic or be mishandled

`l2BlockNumber()` is returned as a Solidity `uint256`, while the execution-layer RPC APIs use
`u64`. Direct conversion or implicit truncation can panic or produce an invalid lookup.

A malicious or malformed game should not be able to terminate the challenger process.

### 2. The boolean `is_invalid` model cannot represent an unavailable validation result

A boolean `is_invalid` model is sufficient only when validation always produces a definitive answer.
That is not true for a claim whose L2 block number is beyond the tip currently visible to the
challenger. The claim may be maliciously ahead of the historical safe head, or the challenger's
rollup/execution nodes may simply be behind and unable to evaluate it yet.

A failed or incomplete lookup is therefore neither proof that the claim is valid nor proof that it
is invalid. It must be represented explicitly as a third state rather than being silently treated
as valid or permanently skipped.

The challenger should reliably distinguish:

- `Valid`: the claim was verified;
- `Invalid`: the claim was deterministically disproven;
- `Unavailable`: the claim cannot currently be evaluated.

Treating `Unavailable` as either `Valid` or a permanently skipped game can allow an invalid claim to
remain unchallenged.

### 3. One unavailable game can block the entire sync loop

A game may be temporarily unqueryable because:

- the execution node is behind;
- historical state is unavailable or pruned;
- the RPC request fails;
- the rollup node has not processed the relevant L1 block yet.

The challenger should isolate this game and continue processing later games. The unavailable game
must remain eligible for retry in a subsequent sync cycle.

### 4. Historical SafeDB data is needed to distinguish an invalid claim from a lagging challenger

A game contains the L1 head associated with its creation. Claim validation should use the local
safe head corresponding to that historical L1 view. The historical SafeDB query is important for
distinguishing two otherwise similar cases:

- If the claimed L2 height is above the local safe head recorded for the game's L1 view, the claim
[...]

PR #982, opened 2026-09-02
fix(challenger): make game validation fail-closed and retryable across sync cycles

KyrinCode:
# [Challenger] Make game validation fail-closed and retryable across sync cycles

Related issue: [#981](https://github.com/succinctlabs/op-succinct/issues/981)

## Summary

This pull request hardens the fault-proof challenger against malformed game data, temporarily
unavailable validation backends, L1 reorg-related state inconsistencies, historical safe-head
ambiguity, and stale transactions.

The dispute game does not require a proposer to prove the claimed L2 height or root at game
creation time. An invalid claim can therefore win without a proof if the challenger cannot evaluate
it and submit a challenge before the challenge window expires.

The changes preserve the existing challenger lifecycle while making validation outcomes, retry
behavior, L1 read consistency, and action preflight explicit.

## Why is this a large change?

The size of this pull request is primarily caused by limitations in the existing validation model
and control flow, rather than by an expansion of the Challenger's product scope.

The previous implementation coupled a boolean `is_invalid` result with game discovery, cached game
state, synchronization errors, and action scheduling. That model cannot safely represent a claim
that is currently unverifiable: a claim beyond the visible node tip may be maliciously ahead of the
historical safe head, or the challenger's own backend may simply be behind or missing history.

The required behavior therefore cannot be implemented reliably through isolated local patches. The
Challenger needs structural changes to:

- represent `Valid`, `Invalid`, and `Unavailable` as distinct outcomes;
- isolate discovery and validation failures while retaining retryable games;
- separate a pinned historical synchronization view from fresh action preflight;
- evaluate claims against the historical SafeDB view associated with each game; and
- separate chain-specific validation from the shared Challenger lifecycle.

The first four commits are correctness and reliability fixes. The fifth commit is an intentionally
separate extensibility refactor: after the validation flow is made explicit, it extracts the
chain-specific validation backend so other teams or deployments can reuse the common lifecycle with
their own validation logic. The existing game lifecycle, parent-game handling, deadlines, and
transaction purposes are preserved.

### 1. Harden L2 output-root lookups

- Safely convert contract-provided `uint256` L2 block numbers before using execution-layer APIs.
- Prevent malformed or unrepresentable heights from panicking the challenger.
- Use header-only L2 reads where possible.
- Handle post-Isthmus `withdrawalsRoot` and pre-Isthmus proof lookup behavior correctly.

### 2. Add an explicit `Unavailable` validation state

Game validation now has three outcomes:

```text
Valid
Invalid(reason)
Unavailable(reason)
```

This distinguishes a deterministically invalid claim from a claim that cannot currently be
evaluated because the challenger backend is behind, missing historical state, or experiencing an
RPC failure.

Unavailable games are retained and retried. Only deterministic `Invalid` validation results
authorize an output-root challenge; the existing parent-loss rule remains an independent reason to
challenge a game.

### 3. Isolate failures per game

- A failed game discovery does not prevent later factory indices from being processed.
- A failed game refresh does not block unrelated challenge, resolution, or bond-claim work.
- Failed discoveries and unavailable validations remain eligible for later retry.
- Stale action flags are cleared when a refresh fails, preventing actions based on old state.
- Metrics are added for unavailable games and isolated synchronization failures.

### 4. Pin synchronization reads and add configurable L1 confirmations

- Add `SYNC_L1_CONFIRMATIONS`, analogous to the proposer confirmation setting.
- Pin factory, game, parent, registry, and credit reads in one synchronization cycle to one
  confirmed canonical L1 block.
- Reject confirmed L1 height regressions relative to the highest accepted snapshot.
- Continue processing an unchanged confirmed height so unavailable games and failed actions can be
  retried.
- Expose the selected confirmed head and confirmation lag through metrics.

For production deployments, operators must explicitly configure a non-zero confirmation depth that
matches the target L1's finality and reorg assumptions. This setting is a safety margin; it is not a
complete cross-cycle reorg detection mechanism.

### 5. Validate claims against the historical SafeDB view

- Resolve each game's `l1Head` to its canonical L1 block.
- Query the op-node SafeDB historical API for the safe head at that historical L1 view.
- Classify a claim above the historical local-safe head as deterministically invalid.
- Classify op-node lag, missing SafeDB history, unavailable execution state, and node mismatches as
  unavailable and retryable.
- Validate L1/L2 chain IDs and the pairing between the op-node and execution node at startup.
- Fail closed and retry startup validation until the configured backend is trustworthy.

This prevents a claim from being considered valid merely because the current execution node has
eventually advanced to the claimed height.

### 6. Add final action preflight

Immediately before broadcasting a challenge, resolution, or bond-claim transaction, the challenger
rechecks the relevant latest canonical state.

Stale or unavailable preflight results skip only the current action. Unconfirmed or transport-
ambiguous transaction results do not clear action state prematurely; the next synchronization cycle
re-reads canonical state before retrying or clearing it.

### 7. Extract chain-specific validation

The shared challenger now consumes an injected `GameValidator` backend. This allows other teams or
deployments to reuse the common Challenger lifecycle while supplying their own claim-validation
logic and data sources.

This commit is not a new protocol correctness fix. It is an extensibility refactor layered on top of
the preceding bug fixes.

The shared lifecycle remains responsible for:

- game discovery;
- state synchronization;
- validation retry policy;
- parent-game handling;
- deadlines;
- challenge, resolution, and bond-claim actions;
- action preflight.

The validator is responsible for startup backend validation and chain-specific claim validation. The
standard OP Stack binary uses `OPStackGameValidator`.

## Commit-by-commit review order

The commits are intentionally ordered and can be reviewed independently within this pull request.
The first four commits fix Challenger correctness and reliability issues; the fifth commit improves
extensibility:

1. `fix(challenger): harden L2 output root lookups`
   - Protects execution-layer lookups from malformed L2 heights and handles header/proof details.
2. `fix(challenger): isolate unavailable game validation`
   - Introduces the tri-state validation model, per-game isolation, and retry behavior.
3. `fix(challenger): pin L1 reads and preflight actions`
   - Adds configurable L1 confirmations, pinned sync snapshots, regression protection, and action
     preflight.
4. `fix(challenger): validate claims against historical safe heads`
   - Adds historical SafeDB validation, node-pair checks, and startup readiness validation.
5. `refactor(challenger): extract game validation backend`
   - Moves OP Stack-specific validation behind `GameValidator` so other validation backends can
     reuse the shared lifecycle without changing its semantics. This is an extensibility refactor,
     not an additional bug fix.

## Behavioral invariants

- `Unavailable` is never treated as `Valid`.
- `Unavailable` is never permanently discarded while the game remains actionable.
- `Unavailable` never authorizes an output-root challenge.
- Only deterministic `Invalid` validation results authorize an output-root challenge; the existing
  parent-loss challenge rule remains unchanged.
- A single unavailable or malformed game cannot starve later games.
- Synchronization reads within one cycle use one confirmed L1 snapshot.
- Action transactions are preceded by a fresh eligibility check.
- Uncertain transaction outcomes are re-evaluated rather than converted into premature state
  transitions.
- Existing parent-loss, resolution, and bond-claim lifecycle behavior is preserved.

## Testing

The commits add or update tests covering:

- L2 block numbers at and above `u64::MAX`;
- header and pre-Isthmus proof lookup behavior;
- valid, invalid, and unavailable validation outcomes;
- isolated game failures and retry behavior;
- discovery cursor progress across failed games;
- pinned L1 confirmation boundaries and confirmed-head regressions;
- stale action preflight and preflight isolation;
- historical SafeDB boundaries and canonical hash checks;
- op-node/execution-node pairing validation;
- validator injection and lifecycle behavior preservation.

## Scope and non-goals

In scope:

- Shared OP Stack challenger correctness and reliability.
- Configurable L1 confirmation depth and deployment documentation.
- Historical SafeDB-based claim validation.
- The `GameValidator` abstraction required to keep chain-specific validation out of the shared
  lifecycle.

Out of scope:

- Changes to the dispute-game contract or lifecycle semantics.
- Implementing additional validation backends beyond the standard OP Stack backend.
- Introducing alternative endpoint failover or permanent-skip semantics for unavailable games.
- Complete cross-cycle L1 reorg detection beyond the confirmation margin and canonical-hash checks
  described above.

## Deployment note

Before production deployment, set `SYNC_L1_CONFIRMATIONS` to a non-zero value selected for the
target L1's finality and expected reorg risk. The default value of `0` is not a production reorg
safety guarantee.

The Challenger must also use one dedicated op-node through `L2_NODE_RPC` and the execution node
paired with it through `L2_RPC`. The op-node must have SafeDB enabled with `--safedb.path`, retain
historical data covering the active challenge window, and expose the historical safe-head API used
by claim validation. The deployment must not use a node pool, `FollowSource`, or `SuperAuthority`
for this node pair, because the Challenger relies on stable historical local-safe semantics.

Both requirements are production gates: confirmations provide an L1 reorg safety margin, while
SafeDB and the correctly paired node backends provide the historical view needed to distinguish an
invalid claim from a temporarily unavailable validation backend.


fault-proof/bin/challenger.rs
```diff
@@ -58,12 +61,19 @@ async fn main() -> Result<()> {
         l1_provider.clone(),
     );
 
-    let mut challenger = OPSuccinctChallenger::new(
+    let game_validator = Arc::new(OPStackGameValidator::new(
+        l1_provider.clone(),
+        ProviderBuilder::default().connect_http(validator_config.l2_rpc),
```
fakedev9999: These providers use Reqwest default client, which has no request timeout. A stalled L1, execution-node, or op-node call can hold the serial sync loop forever, so let's use one finite-timeout client for all three and test that the next game proceeds after timeout.

fault-proof/src/challenger.rs
```diff
@@ -184,175 +195,261 @@ where
...
+                        ?error,
+                        "Failed to retry game discovery; will retry next cycle"
+                    );
+                    ChallengerGauge::SyncFailures.increment(1.0);
+                }
+            }
+        }
+
+        // Discover every new factory index independently. A failed index is retained for retry,
+        // while the discovery cursor continues so later games cannot be starved.
```
fakedev9999, line 227: The current test only exercises record_discovery_attempt and does not run this failure path. Please add a sync_state test where a parent discovery fails, its dependent child schedules no unsafe action, an independent later game proceeds, and the parent succeeds on retry.

book/fault_proofs/challenger.md
```diff
@@ -74,10 +81,47 @@ CHALLENGER_METRICS_PORT=9001          # The port to expose metrics on
 # Testing Configuration (Optional)
 MALICIOUS_CHALLENGE_PERCENTAGE=0.0    # Percentage of valid games to challenge for testing (0.0 = disabled)
 
+# L1 State Snapshot Configuration (Optional)
+SYNC_L1_CONFIRMATIONS=0               # L1 blocks behind latest used for pinned sync reads
```
fakedev9999: This production-facing example leaves SYNC_L1_CONFIRMATIONS=0, while issue #981 requires an explicit non-zero production depth. Can you make that requirement explicit here and state that the selected depth provides a reorg margin, not finality?

KyrinCode:
Received.Kyrin

fault-proof/src/challenger.rs
```diff
@@ -499,7 +659,44 @@ where
...
+        let status = contract.status().block(latest_block).call().await?;
+        let claim_data = contract.claimData().block(latest_block).call().await?;
+        let deadline = U256::from(claim_data.deadline).to::<u64>();
+
+        Ok(challenge_preflight_ready(status, claim_data.status, latest_timestamp, deadline))
+    }
+
+    async fn resolution_preflight(&self, game: &Game) -> Result<bool> {
+        let contract = OPSuccinctFaultDisputeGame::new(game.address, self.l1_provider.clone());
+        Ok(resolution_preflight_ready(contract.status().call().await?))
```
fakedev9999: `status == IN_PROGRESS` is necessary but not sufficient. `resolve()` also requires a resolved parent and, unless the parent lost, `gameOver()`, so please recheck those conditions at one pinned latest block and add a stale-resolution test.

fault-proof/src/challenger.rs
```diff
@@ -499,7 +659,44 @@ where
...
+    }
+
+    async fn claim_preflight(&self, game: &Game, recipient: Address) -> Result<bool> {
+        let contract = OPSuccinctFaultDisputeGame::new(game.address, self.l1_provider.clone());
+        Ok(claim_preflight_ready(contract.credit(recipient).call().await?))
```
fakedev9999: `credit > 0` is necessary but not sufficient. `claimCredit()` calls `closeGame()`, which requires finalization, so please check `isGameFinalized` and credit at one pinned latest block and add a stale-claim test.

fault-proof/src/challenger.rs
```diff
@@ -749,15 +949,517 @@ pub struct Game {
...
+    }
+
+    fn push_active_unchallenged_game(asserter: &Asserter, deadline: u64) {
+        asserter.push_success(&GameStatus::IN_PROGRESS.abi_encode());
+        asserter.push_success(
+            &ClaimData {
+                parentIndex: u32::MAX,
+                counteredBy: Address::ZERO,
+                prover: Address::ZERO,
+                claim: B256::ZERO.into(),
```
fakedev9999: The repository clippy policy fails on this change: this `.into()` is redundant, and `fault-proof/src/lib.rs:300,321` reassign fields after `Default::default()`. Please fix all three warnings and rerun the all-feature, all-target clippy command with `-D warnings`.

fault-proof/src/challenger.rs
fakedev9999: `resolution_preflight` and `claim_preflight` imply complete eligibility checks, but each only queries the mutable stale-state signal. Wdyt of renaming them to `resolution_still_pending` and `claim_credit_available`, inline the one-line `*_ready` helpers, and narrow the public resolution comment and error text to match.

fault-proof/bin/challenger.rs
KyrinCode: All L1, execution-node, and op-node providers now use a finite request timeout via a shared Reqwest client.

book/fault_proofs/challenger.md
KyrinCode: Updated the example to use SYNC_L1_CONFIRMATIONS=5 and documented

fault-proof/src/challenger.rs
KyrinCode: The all-feature, all-target Clippy check now passes with -D warnings.

fault-proof/src/challenger.rs
KyrinCode: Renamed.

fault-proof/src/challenger.rs
KyrinCode: Added the requested sync_state test.

fault-proof/src/op_stack_game_validator.rs
```diff
@@ -0,0 +1,664 @@
...
+
+        let safe_head = self.safe_head_at_l1(game_l1_number).await.map_err(|error| {
+            ChallengerGauge::SafeDbQueryErrors.increment(1.0);
+            classify_safe_db_error(error)
+        })?;
+        self.validate_safe_head_response(game_l1_number, safe_head).await?;
+
+        let second_status = self.op_node_sync_status().await.map_err(rpc_unavailable)?;
+        record_op_node_l1_lag(&second_status);
+        if !op_node_watermarks_are_usable(
```
fakedev9999, line 168: Both samples check `current_l1.number`, but not its hash. An op-node on a stale L1 fork can pass these height checks while SafeDB returns a common-ancestor record whose hashes also pass. This can incorrectly mark a valid claim as `Invalid(ClaimAheadOfLocalSafeHead)`.

Please verify both sampled `current_l1` hashes against `L1_RPC`, return `Unavailable` on mismatch, and add a regression test for this case.

fault-proof/tests/common/process.rs
```diff
@@ -136,22 +137,34 @@ pub async fn new_challenger(
...
         malicious_challenge_percentage: malicious_percentage.unwrap_or(0.0),
+        sync_l1_confirmations: 0,
         tx_confirmation_timeout: 60,
     };
 
     let l1_provider = ProviderBuilder::default().connect_http(rpc_config.l1_rpc.clone());
     let anchor_state_registry =
         AnchorStateRegistry::new(*anchor_state_registry_address, l1_provider.clone());
     let factory = DisputeGameFactory::new(*factory_address, l1_provider.clone());
+    let game_validator = Arc::new(OPStackGameValidator::new(
```
fakedev9999: This fixture starts a fresh Anvil chain and replaces only `L1_RPC`, leaving the external L2 and op-node endpoints unchanged. The production validator rejects this pairing, so `try_init()` keeps retrying before the Challenger tests reach their assertions.

Please inject a test-only `GameValidator` that preserves output-root comparisons and returns `Unavailable` on RPC errors. Cover production startup validation separately using the existing RPC mocks. This fixture issue is separate from the fork-secret CI failure.

fakedev9999 (commented):
Two follow-up comments from the final review.

fault-proof/src/op_stack_game_validator.rs
KyrinCode: Added

fault-proof/tests/common/process.rs
KyrinCode: done

fakedev9999:
Closing in favor of #998 to run CI from the upstream repository. We'll merge there once CI passes. Thanks @KyrinCode!

